PR #116, opened 2014-09-05
drop DynamicRouter::match

### LINKED ISSUE #105: "Dangerous" code in DynamicRouter?
While diving into the Drupal code, I stumbled upon this line https://github.com/symfony-cmf/Routing/blob/master/DynamicRouter.php#L205 where a `Request` is created with only the path info. That looks like a bad idea as everything except the path info falls back to default values.

dbu:
This method is redundant to matchRequest but prone to information loss. See discussion in #105. The ChainRouter still has the match() method.

| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | yes |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #105 |
| License | MIT |
| Doc PR | - |

fix #105 

TODO:
* [ ] UPGRADE.md file

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Changelog
   there was a potential information loss by re-creating the request. If you use
   DynamicRouter directly, get access to the Request object or if you are
   stand-alone create the request with Request::createFromGlobals().
+  Deprecated ChainedRouterInterface as it adds no additional information over
+  VersatileGeneratorInterface.
 
 1.3.0-RC1
 ---------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+* **2014-09-05**: Dropped DynamicRouter::match and DynamicRouter no longer
+  implements RouterInterface but only RequestMatcherInterface and
+  UrlGeneratorInterface. The match method is redundant with matchRequest but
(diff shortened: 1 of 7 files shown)


lsmith77:
@fabpot do you have any plans here?

dbu:
@tobion do you have an idea what symfony 3.0 thinks about router match vs matchRequest? is symfony going to drop one of them or keep them both?

Tobion:
I guess if we start transitioning to PSR-7 in symfony 3.0 we would also more likely support matching against RequestInterface from PSR-7. For everything else, I'm not sure it's worth to break BC so hard just to remove one of the existing.

dbu:
@Tobion well, to me its very confusing that the RouterInterface mandates `match` and then there is a RequestMatcherInterface with a different signature. in the end we do in this bundle re-create a Request when `match` is called so that we have one consistent code. fabien is even afraid this could lead to security issues: https://github.com/symfony-cmf/Routing/issues/105

WouterJ:
@dbu can you please rebase and finish this PR?

dbu:
symfony did not make up their mind about the future of match(). i suggest we leave this for version 2 and see what happens until we do version 3. there is PSR-7 and another psr about server requests, maybe that will replace the symfony request anyways at some point.

ElectricMaxxx:
pls reopen if still interessted.

PR #132, opened 2015-04-06
Dynamic router pre generate event

dbu:
continue #131


dbu:
the build failures are adressed in #133 - they are about php version 5.3 and memory limits, nothing to do with these changes.

thanks a lot @benglass for this contribution!

PR #133, opened 2015-04-06
get rid of the build errors because of memory in php 5.3

dbu:


dbu:
the failure with the 5.3.3 looks like either composer became more picky, or travis changed their setup since we started the minimal requirements. @WouterJ do you have an idea what we can do here?

WouterJ:
@dbu the problem is caused by installing composer using the installer in the before script. Travis doesn't have the openssl extension in PHP 5.3.3: http://docs.travis-ci.com/user/languages/php/#PHP-installation

 > Note that the OpenSSL extension is disabled on php 5.3.3 because of compilation problems with OpenSSL 1.0.

This causes the installer to fail.

You can install composer in the same way as Travis workers do it:

    curl -sS http://getcomposer.org/composer.phar > composer

dbu:
thanks, trying this now. lets see

WouterJ:
Still doesn't work as Composer needs to much time to resolve deps. Maybe we can solve this by adding `-vvv`, but I don't think so...

dbu:
lets see if [20 minutes are enough](http://docs.travis-ci.com/user/build-timeouts/#Build-times-out-because-no-output-was-received)

WouterJ:
It works now!

(however, it's really strange that it takes this long...)

Issue #134, opened 2015-04-20
New cmf_routing.pre_dynamic_generate overrides the route name


luishdez:
Now all my {{ paths(...) }}  try to load 'cmf_routing.pre_dynamic_generate' instead the original name of the route.

I found that the route is being override by the event name

https://github.com/symfony-cmf/Routing/blob/master/DynamicRouter.php#L174

```php
    public function generate($name, $parameters = array(), $referenceType = false)
    {
        if ($this->eventDispatcher) {
            $event = new RouterGenerateEvent($name, $parameters, $referenceType);
            $this->eventDispatcher->dispatch(Events::PRE_DYNAMIC_GENERATE, $event);
            $name = $event->getName(); <------ 
            $parameters = $event->getParameters();
            $referenceType = $event->getReferenceType();
        }
        return $this->getGenerator()->generate($name, $parameters, $referenceType);
    }
```

Note: That getName() method in event for Symfony 3.0 is deprecated so I guess this will work with 3.0

dbu:
oh f... sorry for that one. very stupid, we run into a name clash about
"name"

i think the only sane solution is to call our thing routeName and
setRouteName / getRouteName. luckily there is no release yet of this
version.

can you do a pull request to adjust RouterGenerateEvent and the
DynamicRouter accordingly? if not, i can do that soonish.

luishdez:
Sure no problem! 

Checking the event I see that this is the name of the route or an instance of Route. Shouldn't we call this just `route` ? getRouteName() and getting a route instance doesn't seem very logic.

```php
    /**
     * The name of the route or the Route instance to generate.
     *
     * @var string|Route
     */
    private $name;
```

dbu:
indeed, forgot about that. yes, very good point. lets call this getRoute
then.

PR #135, opened 2015-04-23
Rename event property from name to route to avoid conflicts

luishdez:
This patch fix https://github.com/symfony-cmf/Routing/issues/134

| Q | A |
|--- |--- |
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Tests pass? | yes
Fixed tickets |https://github.com/symfony-cmf/Routing/issues/134 
License | MIT 
Doc PR | -

### DIFF
--- a/DynamicRouter.php
+++ b/DynamicRouter.php
@@ -66,7 +66,8 @@ class DynamicRouter implements RouterInterface, RequestMatcherInterface, Chained
     protected $sortedEnhancers = array();
 
     /**
-     * The regexp pattern that needs to be matched before a dynamic lookup is made
+     * The regexp pattern that needs to be matched before a dynamic lookup is
+     * made
      *
      * @var string
      */
@@ -156,7 +157,7 @@ class DynamicRouter implements RouterInterface, RequestMatcherInterface, Chained
      * If the generator is not able to generate the url, it must throw the
      * RouteNotFoundException as documented below.
      *
-     * @param string|Route $name          The name of the route or the Route instance
+     * @param string|Route $route         The name of the route or the Route instance
      * @param mixed        $parameters    An array of parameters
      * @param bool|string  $referenceType The type of reference to be generated (one of the constants in UrlGeneratorInterface)
      *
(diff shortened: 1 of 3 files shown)


DynamicRouter.php
```diff
@@ -166,17 +167,17 @@ public function getGenerator()
      *
      * @api
      */
-    public function generate($name, $parameters = array(), $referenceType = false)
+    public function generate($route, $parameters = array(), $referenceType = false)
```
dbu: i don't want to rename this here, as its implementing the RouterInterface which calls the thing `$name`

dbu:
thanks for the pull request. i agree with the changes on the event, but not the changes on the DynamicRouter.

@benglass are you already using this feature? if so, you should watch this issue and when we merge, update your code. the nasty thing about it is you will get no error. but setName is a general thing of events, you can change what the event is called. calling setName instead of setRoute will simply have no effect anymore...

dbu:
we don't need to adjust the documentation really, as https://github.com/symfony-cmf/symfony-cmf-docs/pull/672 is not yet merged.

benglass:
@dbu we are not using this yet but thanks very much for the heads up

WouterJ:
This should be mentioned as a BCbreak in the changelog

dbu:
@WouterJ it was added 2 weeks ago and never released. but to be on the safe side we could still add it to CHANGELOG.md, yes

luishdez:
Ops I missed that RouterInterface. 

Regarding the Changelog if you guys are talking about this method rename… I don't see the point. It was broken, so nobody has been able to use it anyways…???

Looks that something is failing here https://travis-ci.org/symfony-cmf/Routing/builds/59834355 but I'm not able to reproduce it… 

unittest

WouterJ:
@luishdez those tests are based on timing and can randomly fail. I've restarted them now and they pass.

dbu:
thanks a lot @luishdez !

@WouterJ i concur with his analysis that this could not have worked in real life (it worked only in the situation of the test cases i think)

PR #136, opened 2015-04-24
[2.2] Route#getPattern is renamed to getPath

WouterJ:
As `~2.2` is set as requirement, this can be done without breaking anything.

After this, the Routing component is deprecation free.


dbu:
awesome!

we could also upgrade the minimum requirement to 2.3 if it helps for something.

PR #139, opened 2015-05-02
Adding hasRouters() to ChainRouter

dragoonis:


dbu:
hi, what is the reason you need this?

dragoonis:
@dbu I need to verify if any routers exist in the ChainRouter before I begin doing things with it.

At the moment I'm forced to do:

``` php
    /**
     * @return bool
     */
    public function hasRouters()
    {
        $routers = $this->sortRouters();
        return !empty($routers);
    }
```

dragoonis:
@dbu I'm going to fix the commits, I tried to achieve this via github.com interface and failed :)

dbu:
ok, i see that. but why do you need to know if the chain router has routers in it and the number of routers it has? not saying this is wrong, but trying to understand.

dragoonis:
@dbu 
My initial use case is to throw an exception if no routes have been registered in thee system.
Secondly, if I can't detect this any routers in the chain, I will have to initiate a matching process on nothing.

ChainRouter.php
```diff
@@ -306,4 +306,25 @@ public function getRouteCollection()
...
     }
+    
+    /**
+     * Identify if any routers have been added
+     * 
+     * @return bool
+     */
+    public function hasRouters()
+    {
+        return !empty($this->routers);
```
dbu: can we make this `return count($this->routers) > 0;` please?

ChainRouter.php
```diff
@@ -306,4 +306,25 @@ public function getRouteCollection()
...
+    }
+    
+    /**
+     * Get the number of added routers
+     * 
+     * @return integer
+     */
+    public function getNumRouters()
```
lsmith77: don't really see a use case for this method .. or rather it seems too specific to me imho

ChainRouter.php
```diff
@@ -306,4 +306,25 @@ public function getRouteCollection()
...
+    {
+        return count($this->routers);
```
dbu: do we really need this? i would suggest you do `count($chainRouter->all())` for this. the count would only ever be relevant when debugging, and there performance should not be critical. and the sorting gets cached, so when the chainrouter is also used in the request, $chainRouter->all() costs quasi nothing.

ChainRouter.php
dragoonis: well, we don't need PHP to count() the zval, we just need to see if it has data in it - this is why i use empty() for things when I don't want PHP to calculate the count()

ChainRouter.php
dragoonis: Well - hasRouters() is useful, but it still doesn't tell you *how many* there are. So getNumRouters() adds more value than hasRouters() in my opinion since you can do ``` if(getNumRouters() > 0)```

ChainRouter.php
dragoonis: if you have no routers and call ```all()``` then it will run sortRouters(), call ```krsort()``` and ```array_merge()``` unncessarily. 

We don't want to run the processing for ```all()``` we just need to know ```"are there routers?``` that is all.

ChainRouter.php
dbu: alright, lets keep it as is then.

ChainRouter.php
lsmith77: a chain routers with zero routers indeed makes no sense, so I see it as something that one might always check. but the number of routers is only useful when debugging and even there I doubt that it will help in most cases, since you actually then care more about which routers are registered and less about how many have been registered. now maybe for your application its all people care about, but imho its not general enough and easy enough to get around not having.

so -1 on adding this method

ChainRouter.php
dbu: for this you are adding hasRouters. as said, the `all` caches the result, so unless you do have routers in the chainrouter but do not execute the chainrouter, there is about no additional cost in calling `all` repeatedly.

dbu:
so my suggestion is that you do a pull request for hasRouters (as it is now, with the empty) but we do not add a method to count the routers. there is `all` for that which is certainly fast enough for debugging.

dragoonis:
@dbu this makes sense - if can do 
```
if(hasRouters()) { 
    $routers = $chainRouter->all(); 
    $numRouters = count($routers);
}
```

dbu:
yes - though `if ($this->hasRouters()) {` it would be ;-)

do you want the count even in production? but still, it should not cost much, if you will use the chain router to match a request.

dragoonis:
@dbu at this moment, I don't need the count(), i just need to know if there's data or not.
I was just pointing out that if I need to know the count, then I can do that with ```all()```

I will fix the commits now.

dragoonis:
@dbu 
I have wiped the existing commits and added just one commit now (#3494df8)

dbu:
cool, thanks for the contribution! and let us know when the chain router is used in the ppi framework.

dragoonis:
@dbu @lsmith77 ChainRouter is now part of PPI Framework.
The PPI Chain Router now supports the Symfony router and Aura Router.
https://github.com/ppi/framework/tree/develop/src/Router

dbu:
cool! ping us when you need a tagged release with the hasRouters() method.

PR #146, opened 2015-09-14
VersatileGeneratorInterface getRouteDebugMessage() requires an array.

aheinz-sg:
Due to an error in our application, I was seeing 
`PHP message: PHP Catchable fatal error:  Argument 2 passed to Symfony\Cmf\Component\Routing\DynamicRouter::getRouteDebugMessage() must be of the type array, null given, called in /usr/share/nginx/www/sites/sgmarketplace/vendor/symfony-cmf/routing/ChainRouter.php on line 254 and defined in /usr/share/nginx/www/sites/sgmarketplace/vendor/symfony-cmf/routing/DynamicRouter.php on line 365`
but I was unable to diagnose our issue until I got the fatal error out of the way with the attached patch.

### DIFF
--- a/ChainRouter.php
+++ b/ChainRouter.php
@@ -248,10 +248,10 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
         throw new RouteNotFoundException(sprintf('None of the chained routers were able to generate route: %s', $info));
     }
 
-    private function getErrorMessage($name, $router = null, $parameters = null)
+    private function getErrorMessage($name, $router = null, $parameters = array())
     {
         if ($router instanceof VersatileGeneratorInterface) {
-            $displayName = $router->getRouteDebugMessage($name, $parameters);
+            $displayName = $router->getRouteDebugMessage($name, $parameters ?: array());
         } elseif (is_object($name)) {
             $displayName = method_exists($name, '__toString')
                 ? (string) $name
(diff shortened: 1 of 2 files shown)


ChainRouter.php
```diff
@@ -248,10 +248,10 @@ public function generate($name, $parameters = array(), $absolute = false)
         throw new RouteNotFoundException(sprintf('None of the chained routers were able to generate route: %s', $info));
     }
 
-    private function getErrorMessage($name, $router = null, $parameters = null)
+    private function getErrorMessage($name, $router = null, $parameters = array())
```
dbu: while we are at it, you could typehint array for $parameters now

ChainRouter.php
```diff
@@ -248,10 +248,10 @@ public function generate($name, $parameters = array(), $absolute = false)
...
     {
         if ($router instanceof VersatileGeneratorInterface) {
-            $displayName = $router->getRouteDebugMessage($name, $parameters);
+            $displayName = $router->getRouteDebugMessage($name, $parameters ?: array());
```
dbu: no need for this change, $parameters will now always be an array. and this method is private, so there are no BC considerations.

dbu:
thanks for this fix! goes to prove that every line that is not covered by tests contains some bugs :-(

i added a few details in the comments. can you maybe recreate this PR against the 1.1 branch? the bug is there ever since we added the feature.

PR #147, opened 2015-09-17
ChainRouter: matching: can not use Request from Request::create($url)

pk16011990:
Url `http://www.example.com//foo/` is matched as `http://www.example.com/`

Cause Request created by
```php
$requestForMatching = Request::create($url);
```
can not be valid for `$url = '//foo/';` because
```php
$requestForMatching->getPathInfo() === '/';
$requestForMatching->getRequestUri() === '/';
```

### DIFF
--- a/ChainRouter.php
+++ b/ChainRouter.php
@@ -172,17 +172,12 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
     {
         $methodNotAllowed = null;
 
-        $requestForMatching = $request;
         foreach ($this->all() as $router) {
             try {
                 // the request/url match logic is the same as in Symfony/Component/HttpKernel/EventListener/RouterListener.php
                 // matching requests is more powerful than matching URLs only, so try that first
-                if ($router instanceof RequestMatcherInterface) {
-                    if (empty($requestForMatching)) {
-                        $requestForMatching = Request::create($url);
-                    }
-
-                    return $router->matchRequest($requestForMatching);
+                if (null !== $request && $router instanceof RequestMatcherInterface) {
+                    return $router->matchRequest($request);
                 }
                 // every router implements the match method
                 return $router->match($url);


dbu:
this is changing the behaviour of the chain router and thus a BC break. there are failing tests.

i am actually quite surprised that `Request::create('http://www.example.com//foo/')` would not work. that sounds like a bug in Request::create to me. can you open a PR on symfony/symfony with a test for that and ask if its the intended behaviour?

pk16011990:
Yes, it is BC break.

Request in symfony/symfony works fine:
```php
$request = \Symfony\Component\HttpFoundation\Request::create('http://www.example.com//foo/');
$request->getPathInfo() === '//foo/';
```
but ```Request::create($uri, ...)``` has parameter URI not URL.

In ```ChainRouter::doMatch()``` the ```Request::create()``` method is called with URL without protocol and host - eg. ```/foo/```.
If you pass ```//foo/``` it is parsed as ```{protocol}//{host}/{pathInfo}``` that means that ```foo``` is now host instead of pathInfo.

dbu:
ah, i see. not good, that. 

we might be able to use the domain and protocol from the RequestContext ($this->context) if it is available. if it is not, we can use the information from the superglobals for those 2 things. then we can build a proper URL that Request::create() can handle.

sustmi:
I think that @pk16011990 is right with the pull request.

If somebody calls ```$matcher->match()``` it does not make sense to try to create new ```Request``` instance and call ```matchRequest()``` on the wrapped routers; only ```match()``` methods should be called.

If somebody calls ```$matcher->matchRequest()```, that is another story.

Most places that call the ```matchRequest()``` method have the following logic:
```php
 if ($matcher instanceof RequestMatcherInterface) {
     $matcher->matchRequest($request);
 } else {
     $matcher->match($request->getPathInfo());
 }
```

```ChainRouter``` implements ```RequestMatcherInterface``` by design, so the logic should now be its competence.

Imagine that you have a simple matcher that does not implement ```RequestMatcherInterface``` (and it is not wrapped in ```ChainRouter```). Caller would use the ```match()``` method. But if you wrap this simple matcher into ```ChainRouter```, the caller now chooses to call ```matchRequest()``` method which our simple matcher does not implement. The only sane thing to do in order to preserve the behaviour seems to call ```$matcher->match($request->getPathInfo())``` for routers that do not implement the ```RequestMatcherInterface```.

This is exactly what @pk16011990 's pull request does. ```$request``` is ```NULL``` iff ```match()``` was called on ```ChainRouter```. And if request is not ```NULL```, that means ```matchRequest()``` was called on ```ChainRouter``` and ```doMatch()``` does the ```$router instanceof RequestMatcherInterface``` logic.

sustmi:
In my opinion the failing tests (```testMatchWithRequestMatchers``` and ```testMatchWithRequestMatchersNotFound```) are wrong, because they call the ```match()``` method and expect the ```matchRequest()``` method to be called on wrapped routers.
If I wanted the ```matchRequest()``` to be called I would call it and not ```match()```.

dbu:
i am not sure how often match() is even used. in the context of symfony full stack its never used. i was hoping that symfony 3.0 would drop the UrlMatcherInterface which would greatly simplify things here... but it looks not like this is going to happen.

so your argument seems to make sense. @dantleech @lsmith77 what do you think? is this a cricital BC break if we change the logic to propagate match() when match was called? unless we only have a RequestMatcherInterface that is not also a UrlMatcherInterface - in that case we still need to recreate a fake request - but then we should use the request context for that.

lsmith77:
hmm it sounds to me like its a bug fix indeed .. and those by definition break broken behavior.

dbu:
@lsmith77 which bug? the one about //foo/? and which was the bug and what is correct? it seems indeed ambigous or even correct that this is a url without protocol.

sustmi:
Hello. I made a new PR https://github.com/symfony-cmf/Routing/pull/152 that includes fixed tests.

dbu:
see #152 - we want to just fix how the Request is created, for now. and hope that https://github.com/symfony/symfony/issues/9781 gets us rid of `match` alltogether.

PR #148, opened 2015-09-18
Reimplement #146 against 1.1

aheinz-sg:


aheinz-sg:
*facepalm* PR against wrong branch.... be right back

Issue #151, opened 2015-09-29
DynamicRouter can't be used exactly like ChainedRouter


dawehner:
Given the architecture I'd expect that you could switch off ChainedRouter without having any problem,
but it turns out, you require it for ->match() calls

dbu:
applications should only use what the RouterInterface or the UrlMatcherInterface / RequestMatcherInterface provides.

what exactly is the problem? either its a bug of DynamicRouter or ChainedRouter is cleaning up something wobbly that you are doing, so more hiding a problem you already have.

dawehner:
So if you use DynamicRouter directly but your matcher is NestedMatcher it just doesn't work, its a bit sad that DynamicRouter does not also provide the wrapping like ChainedRouter does

dbu:
do you use match or matchRequest? the preferred way to use any of the
routers is to use matchRequest, to prevent information loss. anything
but the URL has to be made up out of thin air with match()

dawehner:
HI.

Well, for normal routing we use ::matchRequest() which works perfect. For some internal validation of user entered path we used to use ::match() but I think its fine to just replace that single call

dawehner:
See https://www.drupal.org/files/issues/2576809-11.patch for the adaptions which needed to be made

dbu:
if i read that diff correctly, the only change related to the different behaviour is this:
```
-      return $router->match($path);
+      $subrequest = Request::create($path);
+      return $router->matchRequest($subrequest);
```

this seems not too heavy to me.

the rest is having to do with assumptions and the different interfaces that may or may not be implemented when you want to be generic.

```
+    if ($this->chainRouter instanceof RequestContextAwareInterface) {
+      $this->chainRouter->getContext();
+    }
+    throw new \RuntimeException();
```
this is missing a return i think.

i don't know the context but i wonder why you do not require a RouterInterface to be injected if you want to decorate a router.
```
+    if ($this->chainRouter instanceof RouterInterface) {
+      return $this->chainRouter->generate($name, $parameters, $referenceType);
+    }
+    throw new \RuntimeException();
```

dbu:
btw, also not #147 and #152 where people want to change how the ChainRouter match vs matchRequest behaves. and even more related things, there is the idea to only have a matchRequest method in symfony 3: https://github.com/symfony/symfony/issues/9781

dbu:
[...]

PR #152, opened 2015-09-30
ChainRouter: propagate match() call instead of mocking Request and calling matchRequest()

sustmi:
Deprecates PR https://github.com/symfony-cmf/Routing/pull/147

### DIFF
--- a/ChainRouter.php
+++ b/ChainRouter.php
@@ -156,10 +156,14 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
     }
 
     /**
-     * Loops through all routers and tries to match the passed request or url.
+     * Loops through all routers and tries to propagate right method (match or matchRequest
+     * based on whether request is provided.
      *
-     * At least the  url must be provided, if a request is additionally provided
-     * the request takes precedence.
+     * In Symfony/Component/HttpKernel/EventListener/RouterListener.php the logic decides
+     * whether to call match() or matchRequest() according to RequestMatcherInterface.
+     * As ChainRouter always implements RequestMatcherInterface, RouterListener will always
+     * call matchRequest() even if ChainRouter wraps router without RequestMatcherInterface.
+     * Therefore we must apply similar logic to the wrapped routers.
      *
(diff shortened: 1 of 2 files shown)


dbu:
thanks for this pull request. i am however afraid that this could lead to unexpected behaviour changes. see for example #151. 

@sustmi @ pk16011990 is the problem you encounter not solved when you use the matchRequest method with a request that you create correctly?

sustmi:
@dbu , OK lets forget about fixing propagating ```match()```.
Still, I insist that the bug is in ```ChainRouter``` and not Symfony.

```UrlMatcherInterface::match()``` takes pathinfo as parameter and not URL.
You should not pass pathinfo into ```Request::create()```, because it expects URI.
According to [RFC-3986](http://tools.ietf.org/html/rfc3986#section-5.4.1) URI ```//g``` means ```g``` is hostname ie. pathinfo is ```/```.

```Request::create()``` is correct, but ```ChainRouter``` is not using it right.

These commits try to fix the behaviour: https://github.com/sustmi/Routing/compare/master...fix-matching-pathinfo?expand=1

lsmith77:
maybe a safer approach would be to set a "dummy host" in this edge case to prevent the ``//`` being misinterpreted?

dbu:
@sustmi okay, sorry for letting this sit around so long. i propose we add handling for the specific case of path starting with `//` and in that case add //host from $_REQUEST if set, //localhost otherwise. that should avoid the edge case, right? can you do a pull request @sustmi ?

sustmi:
I don't understand. If you have Request instance, you don't have to fake it. Or do you really mean $_REQUEST superglobal?

dbu:
i do mean the superglobal. we can also go for always localhost. its just
for the edge case of //. we have to rethink this for 2.0 but in 1.x i dont
want to change this behaviour so fundamentally. hopefully "match(path)"
wont exist anymore for 2.0 (symfony 3)

dbu:
superseeded by #163

PR #149, opened 2015-09-18
Reimplement #146 against 1.1

aheinz-sg:

PR #153, opened 2015-10-09
VersatileGeneratorInterface getRouteDebugMessage() requires an array.

dbu:
fix #146, replace #149

Due to an error in our application, I was seeing 
`PHP message: PHP Catchable fatal error:  Argument 2 passed to Symfony\Cmf\Component\Routing\DynamicRouter::getRouteDebugMessage() must be of the type array, null given, called in /usr/share/nginx/www/sites/sgmarketplace/vendor/symfony-cmf/routing/ChainRouter.php on line 254 and defined in /usr/share/nginx/www/sites/sgmarketplace/vendor/symfony-cmf/routing/DynamicRouter.php on line 365`
but I was unable to diagnose our issue until I got the fatal error out of the way with the attached patch.

### DIFF
--- a/ChainRouter.php
+++ b/ChainRouter.php
@@ -230,7 +230,7 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
             try {
                 return $router->generate($name, $parameters, $absolute);
             } catch (RouteNotFoundException $e) {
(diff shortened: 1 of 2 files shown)


dbu:
thanks a lot @aheinz-sg !

PR #154, opened 2015-10-19
fix url generator signature using the constant for reference type

Tobion:
The `UrlGeneratorInterface::generate` method uses a "referenceType" instead of "absolute" parameter since symfony version 2.2. The important part is to use the constants as the old values of those are deprecated in 2.8: symfony/symfony#16276

I don't know if 1.3 is the correct target branch. If it should be applied somewhere else, please do so.

### DIFF
--- a/Tests/Routing/ContentAwareGeneratorTest.php
+++ b/Tests/Routing/ContentAwareGeneratorTest.php
@@ -24,7 +24,7 @@ class ContentAwareGeneratorTest extends CmfUnitTestCase
     protected $provider;
 
     /**
-     * @var ContentAwareGenerator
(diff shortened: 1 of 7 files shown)


Tests/Routing/ContentAwareGeneratorTest.php
```diff
@@ -24,7 +24,7 @@ class ContentAwareGeneratorTest extends CmfUnitTestCase
     protected $provider;
 
     /**
-     * @var ContentAwareGenerator
+     * @var TestableContentAwareGenerator
```
Tobion: otherwise the IDE complains that a protected method is used publicly

dbu:
thanks a lot!

PR #155, opened 2015-10-30
Provide a RouteEnhancerCollectorTrait for easier use of route enhancers.

timplunkett:
| Q | A |
| --- | --- |
| Bug fix? | [no] |
| New feature? | [yes] |
| BC breaks? | [yes] |
| Deprecations? | [no] |
| Tests pass? | [yes] |
| Fixed tickets | N/A |
| License | MIT |
| Doc PR | N/A |

### DIFF
--- a/DynamicRouter.php
+++ b/DynamicRouter.php
@@ -23,7 +23,6 @@ use Symfony\Component\Routing\RequestContextAwareInterface;
 use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 use Symfony\Component\Routing\Exception\MethodNotAllowedException;
-use Symfony\Cmf\Component\Routing\Enhancer\RouteEnhancerInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Cmf\Component\Routing\Event\Events;
 use Symfony\Cmf\Component\Routing\Event\RouterMatchEvent;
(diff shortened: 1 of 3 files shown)


timplunkett:
The only BC break is the bump to 5.4. I contemplated adding the trait without removing the duplicated code from DynamicMatcher, but thought better of it.

dbu:
is the reason why you want to split this out separation of concerns, or do you see use cases for reusability?

the upcoming 1.4 release is probably the last in the 1.x, then we go to 2.0

i put this PR into the 2.0 milestone, i think that is the right moment to drop php 5.3 support. afaik symfony 3 will have 5.5 as minimum, we can then just go with that as well.

dbu:
@timplunkett can you elaborate why you want to split out this code?

dbu:
ping @timplunkett

dbu:
ping @timplunkett

timplunkett:
My apologies for not responding! I use GH about once a year so I didn't notice.

Here is the use case for a Drupal module:

I have multiple routes for the same path. Each has "selection conditions" that can be configured in the admin UI. Example conditions are role (auth/anon), current language, type of page being viewed (article, event, etc).

Some of these conditions rely on the request attributes. Specifically they rely on a route enhancer having run, so that on the request path of `/node/5` with the route name `/node/{node}`, `$request->attributes->get('node')` will return the equivalent of `Node::load(5)`.

There is a route filter that needs to consult each candidate route about whether all selection conditions are satisfied.

[This method](http://cgit.drupalcode.org/page_manager/tree/src/Routing/VariantRouteFilter.php#n203) is an approximation of `\Symfony\Component\Routing\Matcher\RequestMatcherInterface::matchRequest` but without re-running route filters.
It is called [here](http://cgit.drupalcode.org/page_manager/tree/src/Routing/VariantRouteFilter.php#n132), the request stack is manipulated, etc.


TL;DR  I have a need to run route enhancers without running route filters.

dbu:
thank you for this explanation, then it makes sense. as we bumped to php 5.6 there is also no problem anymore with using traits... i recreated this PR from the latest master in #191 - can you please have a look at that one and tell if it works for you?

PR #156, opened 2015-11-07
Trigger deprecation notice

WouterJ:


dbu:
build is now failing because of our deprecation notice... how can we not fail because of it? we should still test it while its there

WouterJ:
@dbu by marking the tests as legacy (either by prefixing the method name with `testLegacy` instead of `test` or the class with `Legacy....Test` or by putting them in the `legacy` group).

PR #159, opened 2015-12-06
Added styleci

dantleech:
As discussed on the mailing list.

Once we have ageed on the style configuration we can proceed to add this to all CMF repositories.

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,0 +1,1 @@
+preset: symfony
--- a/ContentAwareGenerator.php
+++ b/ContentAwareGenerator.php
@@ -37,16 +37,16 @@ class ContentAwareGenerator extends ProviderBasedGenerator
 
     /**
      * The content repository used to find content by it's id
-     * This can be used to specify a parameter content_id when generating urls
+     * This can be used to specify a parameter content_id when generating urls.
      *
      * This is optional and might not be initialized.
      *
-     * @var  ContentRepositoryInterface
+     * @var ContentRepositoryInterface
      */
     protected $contentRepository;
 
     /**
-     * Set an optional content repository to find content by ids
+     * Set an optional content repository to find content by ids.
      *
      * @param ContentRepositoryInterface $contentRepository
      */
@@ -56,7 +56,7 @@ class ContentAwareGenerator extends ProviderBasedGenerator
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      *
      * @param string $name       ignored.
      * @param array  $parameters must either contain the field 'route' with a
@@ -76,7 +76,7 @@ class ContentAwareGenerator extends ProviderBasedGenerator
             $route = $this->getRouteByContent($name, $parameters);
         }
 
-        if (! $route instanceof SymfonyRoute) {
+        if (!$route instanceof SymfonyRoute) {
             $hint = is_object($route) ? get_class($route) : gettype($route);
             throw new RouteNotFoundException('Route of this document is not an instance of Symfony\Component\Routing\Route but: '.$hint);
         }
@@ -87,7 +87,7 @@ class ContentAwareGenerator extends ProviderBasedGenerator
     }
 
     /**
-     * Get the route by a string name
+     * Get the route by a string name.
      *
      * @param string $route
      * @param array  $parameters
@@ -100,7 +100,7 @@ class ContentAwareGenerator extends ProviderBasedGenerator
     {
         $route = $this->provider->getRouteByName($name);
         if (empty($route)) {
-            throw new RouteNotFoundException('No route found for name: ' . $name);
+            throw new RouteNotFoundException('No route found for name: '.$name);
         }
 
         return $this->getBestLocaleRoute($route, $parameters);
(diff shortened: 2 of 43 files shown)


.styleci.yml
```diff
@@ -0,0 +1,8 @@
+enabled:
+
+disabled:
+  - align_double_arrow
```
WouterJ: I think we should align double arrows, makes things much more readable imo.

ContentAwareGenerator.php
```diff
@@ -155,9 +155,9 @@ protected function getBestLocaleRoute(SymfonyRoute $route, $parameters)
      * @param array $parameters which should contain a content field containing
      *                          a RouteReferrersReadInterface object
      *
-     * @return SymfonyRoute the route instance
-     *
      * @throws RouteNotFoundException if no route can be determined
+     *
+     * @return SymfonyRoute the route instance
```
WouterJ: not sure I like this ordening and afaik, Symfony core places throws as last as well.

PagedRouteProviderInterface.php
```diff
@@ -1,4 +1,14 @@
...
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2015 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
```
WouterJ: seems wrong (duplicate license header)

.styleci.yml
dbu: but leads to lots of unnecessary diffs when changing code. does symfony code style say something about this? i would stick to that so we can avoid arguing here :-)

ContentAwareGenerator.php
dbu: that seems wrong to me as well. i think just about every code i used had the param / return / throws order.

dbu:
something goes wrong with php 5.3 and the deprecation handler. maybe we should simply drop 5.3 support?

.styleci.yml
WouterJ: Symfony did it before but has changed it a year ago or so, to avoid those big diffs.

In the docs, we still use it as docs are about readability.

dbu: so the way consistent with symfony would be to leave this setting as it currently is, correct?

dantleech: but I aggree with [this post](http://mnapoli.fr/approaching-coding-style-rationally/) when it comes to whitespace alignment

ContentAwareGenerator.php
dantleech: Will try and either remove the CS fix or maybe there is a way to specify
the order.

dantleech: I think having consistency is better than what we are used to, so I would vote to keep this fixer. But we can disable it by disabling `phpdoc_order`.

PagedRouteProviderInterface.php
dantleech: The original had a duplicated header already, probably why we got an extra one here.

dantleech:
Disabled PHP5.3 -- @WouterJ @lsmith77 do you agree with this? or should we keep supporting it?

dantleech:
Also created a script to sync all of the repositories with a single `.styleci.yml`: https://github.com/dantleech/cmf-utils/blob/master/scripts/sync_styleci

WouterJ:
I don't think we should bump the PHP requirement in CMF 1.3. Let's do that in CMF 2.0, which will also drop Symfony 2.3 support and should be released somewhere next year.

dantleech:
In that case we seem to have some problem with the testing component and PHP5.3 I think.

WouterJ:
It's a bug with the PHPunit bridge of Symfony. It has been fixed already: https://github.com/symfony/symfony/commit/a4aafca33d0f6303ff01c8774408c97a4de9ee2f

dantleech:
Readded PHP 5.3 support. Are we otherwise happy with the style guidelines? Shall we leave the docblock `@annotation` ordering as per the `phpdoc_order` fixer?

dbu:
i would skip the phpdoc_order if we can't tell it which way we want things. param / throw / return is a really weird order.

looks like the php 7 travis box is broken again?

.styleci.yml
lsmith77: I would recommend to just use the ``symfony`` preset: https://github.com/doctrine/DoctrinePHPCRBundle/blob/master/.styleci.yml

dantleech:
Yeah, it makes most sense I have updated to use that, although I notice that Symfony itself doesn't currently use that.

The biggest change now is that it uses whitespace to align docblocks, which I am against, but whatever.

lsmith77:
yeah .. I am also not fan of aligning via whitespace .. but oh well.

lsmith77:
looks like the travis php7 phpunit setup is broken:
```
0.04s$ phpunit --coverage-text
/home/travis/.phpenv/versions/7.0/bin/phpunit: line 1: html: No such file or directory
/home/travis/.phpenv/versions/7.0/bin/phpunit: phpunit: line 2: syntax error near unexpected token `<'
```

lsmith77:
retriggeree the php7 builds. they now pass

PagedRouteProviderInterface.php
lsmith77: this still needs to be fixed

WouterJ:
:+1: (I like aligning btw)

PagedRouteProviderInterface.php
dantleech: Fixed.

dantleech:
If the tests pass and we merge this I will endeavour to create PRs for all the other CMF repositories.

PagedRouteCollection.php
```diff
@@ -92,7 +85,7 @@ public function next()
         if (false === $result) {
             $this->loadNextElements($this->current + 1);
         }
-        $this->current++;
+        ++$this->current;
```
dbu: wtf but whatever. it does exactly the same here, as we don't do anything else than increase the value.

dbu:
thanks a lot, dan! looks like there was a lot to fix.

PR #162, opened 2015-12-31
Allow FosJsRouting dev for Symfony 3

WouterJ:


WouterJ:
Build is fixed again

composer.json
```diff
@@ -21,7 +21,8 @@
         "symfony-cmf/testing": "^1.3@dev",
         "symfony/dependency-injection": "^2.0.5|3.*",
         "symfony/config": "^2.2|3.*",
-        "symfony/event-dispatcher": "^2.1|3.*"
+        "symfony/event-dispatcher": "^2.1|3.*",
+        "friendsofsymfony/jsrouting-bundle": "@dev"
```
dbu: do we require this or is it some other bundle that requires it? then the `@dev` should go in there rather than here, i think.

WouterJ: Testing requires it. `@dev` isn't needed, only to support Symfony 3 as a Symfony 3 supporting version has not been released.

However, as minimum stability is now set to stable for all CMF repositories, we need to explicitely allow FOSJsRoutingBundle to be installed in dev when needed.

dbu: okay, i see. i hope to release this library really soon - should we do so with this line or should we try to tag a stable version of jsrouting?

dbu: ah its only a dev dependency so no big deal!

WouterJ: Yeah and it's a dep that's not even used by this library. This is one of the problems of Testing 1.0, eventhough this is a component tested without framework integration, many bundles are still automatically loaded by Testing. For Testing 2.0, I would like to remove all config/required bundles from the component, so each library can make choices themselves.

PR #163, opened 2015-12-31
use the request context to rebuild the request on matching

dbu:
this is an alternative to #147 / #152 without BC breaks.

/cc @pk16011990 @sustmi would that solve your situation? if you set the request context on the router, you will no longer run into the problem with the double leading // making symfony thing the path is the domain.

### DIFF
--- a/ChainRouter.php
+++ b/ChainRouter.php
@@ -180,11 +180,12 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
                 // matching requests is more powerful than matching URLs only, so try that first
                 if ($router instanceof RequestMatcherInterface) {
                     if (empty($requestForMatching)) {
-                        $requestForMatching = Request::create($url);
+                        $requestForMatching = $this->rebuildRequest($url);
                     }
 
                     return $router->matchRequest($requestForMatching);
                 }
+
                 // every router implements the match method
                 return $router->match($url);
             } catch (ResourceNotFoundException $e) {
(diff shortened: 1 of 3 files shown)


ChainRouter.php
```diff
@@ -249,6 +250,46 @@ public function generate($name, $parameters = array(), $absolute = UrlGeneratorI
...
+            $server['SERVER_NAME'] = $this->context->getHost();
+            $server['HTTP_HOST'] = $this->context->getHost();
+        }
+        if ('https' === $this->context->getScheme()) {
+            $server['HTTPS'] = 'on';
+            $server['SERVER_PORT'] = $this->context->getHttpsPort();
+            if (443 !== $this->context->getHttpsPort()) {
+                // this is parsed from the host by symfony request
+                // https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L971
+                $server['HTTP_HOST'] .= ':'.$this->context->getHttpsPort();
```
dbu: @tobion this surprised me - there are really many interactions in the Request object with priorities between different headers and options to set things explicitly...

do you see things i am getting wrong here? like maybe problems with the base path? is there somewhere something already existing to recreate a Request from RequestContext? i only found the other way round, RequestContext::fromRequest.

lsmith77:
build is failing due to the ``friendsofsymfony/jsrouting-bundle`` dependency on Symfony 3.0 which I do not understand since it seems to be allowed:
https://github.com/FriendsOfSymfony/FOSJsRoutingBundle/blob/master/composer.json#L20

I guess its due to the CMF testing bundle not allowing a dev version?
https://github.com/symfony-cmf/Testing/blob/master/composer.json#L24

So maybe using an alias instead of ``@dev`` could work ..?

dbu:
wouter did this in #162 and things looked good. squashed, this should now be green.

only worry i have left is whether the reconstruction is complete, notably whether i have a problem if symfony does not run at `/` of the web server.

lsmith77:
needs another rebase?

Tests/Routing/ChainRouterTest.php
```diff
@@ -309,6 +309,63 @@ public function testMatchWithRequestMatchers()
...
+                ;
+            }))
+            ->will($this->throwException(new \Symfony\Component\Routing\Exception\ResourceNotFoundException()))
+        ;
+        $low
+            ->expects($this->once())
+            ->method('match')
+            ->with($url)
+            ->will($this->returnValue(array('test')))
+        ;
```
WouterJ: This (and the one above) is combining both a mock (expectation of number of calls) and a stub (returning a value). This is considered a bad practice, a method often is either a command (in which a mock should be used) or a query (in which a stub should be used). It makes no sense to verify the amount of calls to a query method, as only the return value is what matters.

Short story: Use `->expects($this->any())` here as you don't care about the amount of calls.

WouterJ:
To me, this looks like some complex logic that doesn't really belong here (but instead, it should maybe be added as a method in `RequestContext`). However, if it solves the problem for now, I'm :+1: (it can easily be refactored to another method without breaking BC).

Tests/Routing/ChainRouterTest.php
dbu: not sure. if we repeatedly call match, this could be a severe performance issue. this is why i want to notice if its called too often.

dbu:
added support for the base path. as wouter says, rebuilding `Request` from `RequestContext` does not really belong here. honestly, i find its quite a hackery, as Request is not just a value object but also handling all the parsing of the PHP values to determine the settings. but i don't see a better option without changes to the Request class.

@lsmith77 wdyt?
@pk16011990 @sustmi would that solve your use case?
@tobion still glad for your input, if you have some time.

lsmith77:
since its in a private method .. I guess we have all options open for a later refactoring.

dbu:
yep. so shall we merge? or wait a bit longer for feedback?

WouterJ:
If tests pass with the case reported in #147, I think it's safe to merge this one. (also, I would like to see some RC1's this weekend for at least Routing stuff, MenuBundle, ContentBundle)

PR #167, opened 2016-01-10
fix the edge case of double leading slash when rebuilding request

### LINKED ISSUE #166: ChainRouter does not handle URLs starting with two slashes properly
This issue follows rejected pull requests: https://github.com/symfony-cmf/Routing/pull/147 and https://github.com/symfony-cmf/Routing/pull/152
and is still not fixed even after merging: https://github.com/symfony-cmf/Routing/pull/163 (sorry for not commenting earlier).

If you pass `//test` as an `$url` parameter into `ChainRouter::match()` method and you have `RequestMatcherInterface` in the chain, it tries to match it as pathinfo `/` with host `test` instead of matching as pathinfo `//test`.

See this failing test: https://github.com/symfony-cmf/Routing/compare/master...sustmi:test-for-matching-double-slash?expand=1

dbu:
fix #166

@sustmi you are right. as it turns out, even specifying the host explicitly, `Request::create('//test')` considers the path as a domain. i now rebuild the full URL, i do hope this now works.

note that you will need to do setContext with the RequestContext for this to work. if there is no request context, we don't do anything more than before the fix.

### DIFF
--- a/ChainRouter.php
+++ b/ChainRouter.php
@@ -141,9 +141,9 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
      *
      * Note: You should use matchRequest if you can.
      */
-    public function match($url)
+    public function match($pathinfo)
     {
-        return $this->doMatch($url);
+        return $this->doMatch($pathinfo);
     }
 
     /**
@@ -162,14 +162,14 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
      * At least the  url must be provided, if a request is additionally provided
      * the request takes precedence.
      *
-     * @param string  $url
+     * @param string  $pathinfo
      * @param Request $request
      *
      * @return array An array of parameters
      *
      * @throws ResourceNotFoundException If no router matched.
      */
-    private function doMatch($url, Request $request = null)
+    private function doMatch($pathinfo, Request $request = null)
     {
         $methodNotAllowed = null;
 
@@ -180,14 +180,14 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
                 // matching requests is more powerful than matching URLs only, so try that first
                 if ($router instanceof RequestMatcherInterface) {
                     if (empty($requestForMatching)) {
-                        $requestForMatching = $this->rebuildRequest($url);
+                        $requestForMatching = $this->rebuildRequest($pathinfo);
                     }
 
                     return $router->matchRequest($requestForMatching);
                 }
 
                 // every router implements the match method
-                return $router->match($url);
+                return $router->match($pathinfo);
             } catch (ResourceNotFoundException $e) {
                 if ($this->logger) {
                     $this->logger->debug('Router '.get_class($router).' was not able to match, message "'.$e->getMessage().'"');
@@ -203,7 +203,7 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
 
         $info = $request
             ? "this request\n$request"
-            : "url '$url'";
+            : "url '$pathinfo'";
         throw $methodNotAllowed ?: new ResourceNotFoundException("None of the routers in the chain matched $info");
     }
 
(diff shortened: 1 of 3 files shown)


ChainRouter.php
```diff
@@ -291,8 +291,19 @@ private function rebuildRequest($uri)
                 $server['HTTP_HOST'] .= ':'.$this->context->getHttpPort();
             }
         }
+        // use the array field to have both domain and port.
+        if (isset($server['HTTP_HOST'])) {
+            $pathinfo = $server['HTTP_HOST'].$pathinfo;
+        } else {
+            $pathinfo = 'localhost'.$pathinfo;
```
dbu: do we want to do this? and if we do, should we do the same for the case without RequestContext being set at all?

sustmi:
Hi @dbu . I tried the fix and it works for me.

ChainRouter.php
sustmi: I hope the question is not on me. I cannot comment on the code because I do not understand it.
I try to keep away from anything that comes into contact with `preparePathInfo()` and `prepareBaseUrl()` methods of `Request`, because it's a black magic.

dbu: this is more a question to @lsmith77

lsmith77: I am not deep enough into this PR to really tell either .. @tobion might know best in general as he knows the entire routing topic really well ..

ChainRouter.php
```diff
@@ -255,26 +255,26 @@ public function generate($name, $parameters = array(), $absolute = UrlGeneratorI
...
         }
 
         $server = array();
-        if ($this->context->getHost()) {
-            $server['SERVER_NAME'] = $this->context->getHost();
-            $server['HTTP_HOST'] = $this->context->getHost();
-        }
         if ($this->context->getBaseUrl()) {
-            $uri = $this->context->getBaseUrl().$uri;
+            $pathinfo = $this->context->getBaseUrl().$pathinfo;
```
Tobion: pathinfo is the path without the base url according to CGI spec. so the variable naming reuse is confusing.

ChainRouter.php
```diff
@@ -291,8 +291,19 @@ private function rebuildRequest($uri)
...
             }
         }
+        // use the array field to have both domain and port.
+        if (isset($server['HTTP_HOST'])) {
+            $pathinfo = $server['HTTP_HOST'].$pathinfo;
+        } else {
+            $pathinfo = 'localhost'.$pathinfo;
+        }
+        if ('https' === $this->context->getScheme()) {
+            $pathinfo = 'https://'.$pathinfo;
```
Tobion: same, clearly not the $pathinfo

ChainRouter.php
```diff
@@ -255,26 +255,26 @@ public function generate($name, $parameters = array(), $absolute = UrlGeneratorI
...
-        }
         if ($this->context->getBaseUrl()) {
-            $uri = $this->context->getBaseUrl().$uri;
+            $pathinfo = $this->context->getBaseUrl().$pathinfo;
             $server['SCRIPT_FILENAME'] = $this->context->getBaseUrl();
             $server['PHP_SELF'] = $this->context->getBaseUrl();
         }
+        if ($this->context->getHost()) {
+            $server['SERVER_NAME'] = $this->context->getHost();
+            $server['HTTP_HOST'] = $this->context->getHost();
```
Tobion: IMO it's easier and cleaner to really build a full URI and then pass this to Request::create. Then you do not need to deal with the server variables at all. The request method is doing that itself: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L330

ChainRouter.php
```diff
@@ -255,26 +255,26 @@ public function generate($name, $parameters = array(), $absolute = UrlGeneratorI
...
         $server = array();
-        if ($this->context->getHost()) {
-            $server['SERVER_NAME'] = $this->context->getHost();
-            $server['HTTP_HOST'] = $this->context->getHost();
-        }
         if ($this->context->getBaseUrl()) {
-            $uri = $this->context->getBaseUrl().$uri;
+            $pathinfo = $this->context->getBaseUrl().$pathinfo;
             $server['SCRIPT_FILENAME'] = $this->context->getBaseUrl();
             $server['PHP_SELF'] = $this->context->getBaseUrl();
```
Tobion: the base url server variables are probably the only ones that you need to deal with as done here because that cannot be deducated from the URI alone.

Tobion:
The query string handling seems to be missing (which is not that important because it's not used by the routing itself AFAIK, but for consistency).

dbu:
thanks a lot for the feedback. indeed i do not need all that header
stuff. i just pushed an updated version.

> The query string handling seems to be missing

i don't understand what you mean here. we just pass around the pathinfo
that was given to us to match. is that without query parameters? or are
the query parameters only handled after the route has been matched?
should i simply append context->getQueryString() to the uri?

Tobion:
Yes, the pathinfo does not contain the query string. Just add the query string at the end. See https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L1067

Tobion:
Otherwise looks good now :+1:

dbu:
i think contrary to other parts, the query parameters are actually used in the `create` method. but i added them to the url nontheless for completeness. will merge when its green.

as we discussed yesterday, it would be really nice to have a stand alone Request class cleanly separated from HTTP header and other black magic, with a separate RequestBuilder to build request instances from HTTP. i hope for symfony 4 :-)

ChainRouter.php
```diff
@@ -255,42 +255,32 @@ public function generate($name, $parameters = array(), $absolute = UrlGeneratorI
...
-                $server['HTTP_HOST'] .= ':'.$this->context->getHttpPort();
-            }
+        $host = $this->context->getHost() ?: 'localhost';
+        if ('https' === $this->context->getScheme() && 443 !== $this->context->getHttpsPort()) {
+            $host .= ':'.$this->context->getHttpsPort();
+        }
+        if ('http' === $this->context->getScheme() && 80 !== $this->context->getHttpPort()) {
+            $host .= ':'.$this->context->getHttpPort();
         }
+        $uri = $this->context->getScheme().'://'.$host.$uri.$this->context->getQueryString();
```
Tobion: this does not work. it's missing the `?` before the query string. this is why i linked the exmaple implementation https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L1067

dbu: argh. thanks for spotting. i fixed it in the master branch as i already
merged this.

Issue #171, opened 2016-04-03
Create content_on_method enhancer


ElectricMaxxx:
To map REST routes (and other use cases) we need an enhancer for to map on controller methods based on the HTTP method. 

The related discussion can be found: [here](https://github.com/symfony-cmf/ContentBundle/issues/93) There is also an example implementation mentioned or can be found [here](https://github.com/ElectricMaxxx/RestBundle)

The enhancer should be available in the bundle too.

dbu:
we need to find a better name. `content_on_method` would suggest that
you add different content based on the request method.

but what you want to do, if i get it correctly, is an enhancer that
selects the controller based on the content class and the method. this
sounds like being able to have more than one controllers_by_class,
limited to methods.

the normal controllers_by_class is handled by
[FieldByClassEnhancer](https://github.com/symfony-cmf/Routing/blob/master/Enhancer/FieldByClassEnhancer.php).
we could create a MethodConditionalEnhancer that wraps another enhancer
and only triggers that if the request method is in the list of methods
that was specified.

for the bundle configuration, we would have to think of a way to add
methods to controller_by_class and have several controller_by_class.

ElectricMaxxx:
That sounds like a plan. 

In general this enhancer would easily add the [method]Action to a given controller as a a service key. The _content isn't touched and should be set on the usual way. So controller_by_class mapping sounds right. 

Would have one problem: doing it they way described above i would expect to have controllers implementing all methods - a kind of a CRUD (Rest) all in one solution. But i think it would he better to expect a controller implementing the methods the user whants to serve. Means implement Interfaces devided by methods (RestPostAwareController, RestDeleteAwareController, ...). implenting the Controller Class and checking the interface in the enhancer is to expensive for the routing - i think. You have got any idea for that?

dbu:
i would not do any magic. if people want magic method names, they
hopefully use FOSRestBundle which does that nicely.

i would simply add that mapper, and allow explicit configuration of
controller_by_class that is limited by method. then people can map
however they need to.

ElectricMaxxx:
Doing so i would add some configuration options combined with Service definitions based on field_by_class enhancer in the bundle only? I would lose the the http method Information. 

But that would mean to configure each http method i whant to Support for each content Class i what to support.

ElectricMaxxx:
I could add a list of controller:[method]Action pairs per content class.

dbu:
i would make the configuration something like

```yaml
controller_by_class:
[...]

PR #172, opened 2016-05-02
[POC] let enhancer work with http method aware defaults

ElectricMaxxx:
Try to let the enhancers work in a HTTP method aware manner as discussed in #171

### DIFF
--- a/DependencyInjection/Compiler/RegisterRouteEnhancersPass.php
+++ b/DependencyInjection/Compiler/RegisterRouteEnhancersPass.php
@@ -27,23 +27,23 @@ class RegisterRouteEnhancersPass implements CompilerPassInterface
     /**
      * @var string
      */
-    protected $dynamicRouterService;
+    protected $conditionalEnhancer;
 
     protected $enhancerTag;
 
-    public function __construct($dynamicRouterService = 'cmf_routing.dynamic_router', $enhancerTag = 'dynamic_router_route_enhancer')
+    public function __construct($conditionalEnhancer = 'cmf_routing.conditional_enhancer', $enhancerTag = 'dynamic_router_route_enhancer')
     {
-        $this->dynamicRouterService = $dynamicRouterService;
+        $this->conditionalEnhancer = $conditionalEnhancer;
         $this->enhancerTag = $enhancerTag;
     }
 
     public function process(ContainerBuilder $container)
     {
-        if (!$container->hasDefinition($this->dynamicRouterService)) {
+        if (!$container->hasDefinition($this->conditionalEnhancer)) {
             return;
         }
 
-        $router = $container->getDefinition($this->dynamicRouterService);
+        $router = $container->getDefinition($this->conditionalEnhancer);
 
         foreach ($container->findTaggedServiceIds($this->enhancerTag) as $id => $attributes) {
             $priority = isset($attributes[0]['priority']) ? $attributes[0]['priority'] : 0;
--- a/DynamicRouter.php
+++ b/DynamicRouter.php
@@ -54,18 +54,6 @@ class DynamicRouter implements RouterInterface, RequestMatcherInterface, Chained
     protected $eventDispatcher;
 
     /**
-     * @var RouteEnhancerInterface[]
-     */
-    protected $enhancers = array();
-
-    /**
-     * Cached sorted list of enhancers.
-     *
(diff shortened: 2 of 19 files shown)


ElectricMaxxx:
Failing test looks random, but happened the second time tonight.

ElectricMaxxx:
Maybe i should create a separate enhancer that works before `FieldByClass` to normalize the defaults value that `FieldByClass` can work the way it used to work before. Means to create `"My/Class" => "Service:method"

dbu:
i still like the idea i proposed: a conditional enhancer that can use
different enhancers based on the http method. instead of configuration
what field to update, it would receive a list of method => enhancer
instance and forward to the right enhancer. with one method being
_default to handle fallback if wanted.

that would keep the FieldByClass enhancer as it is before and thus much
simpler.

ElectricMaxxx:
keeping the `FieldByClass` enhancer clean is my idea (last comment) too. But as usual: you have got great ideas, but i don't understand them on the first time. :-)

I try to refrase it:
You mean that conditional enhancer should wrap all others. No HTTP method would mean that the others can to their work in the chain as they used to do. With an HTTP method (i.e. POST) we need to transform the configuration like [that](https://github.com/symfony-cmf/routing/issues/171#issuecomment-206436290) a version the former enhancers still understand like 

...
controllers_by_class:
     My\Class: "service:method"
```
(i would pass it as an configuratio, but the enhancer understands an array of class => service:action)
Doing that transformation would keep the other enhancers as they are. Is it that what you mean?

WouterJ:
> Failing test looks random, but happened the second time tonight.

See #173

ElectricMaxxx:
@dbu slowly i get your idea, but wouldn't it change the current call behavior and the extension class at all? Currently we are passing the configurations directly into parameters and pass them to the enhancers while their service configuration. What you plan is a thing between the ExtensionClass and the DynamicRouter, which should prepare the configuration values. That conditional enhancer can be a kind of a chain enhancer, which is called by the dynamic router, which doesn't recognize the change. But don't you think this should be done in a separate Issue/PR?

dbu:
in this PR we do not need to worry about configuration of the
RoutingBundle (but anything we can express in objects here will be
possible to model with configuration)

my idea is to have an enhancer that does not do anything itself except
determine which enhancer to call, based on the request method. then we
can leave the current enhancers alone. it makes no sense to merge this
pull request first and then doing it later, because we would revert the
changes of this PR which is unnecessarily complicated.

ElectricMaxxx:
@dbu but the enhancers expect some parameters we have to be aware of. Even a conditional Enhancer need those. An so the conditional enhancers need the Information like POST goes on service:postAction to feed the FieldByClass enhancer, that does its job as it used to do. I am completely with you by keeping the enhancers clean as they are. But i would not do some kind of method => enhancer mapping in the code. I think we should pass the name while its Service Definition (i.e. as tag attribute) and ask later isYourName('controller_by_class') while looping through them all. A so called ConditionalEnhancer would do some Kind of normalizing, when the parameters are Method aware. Doing so the enhancers stay, their service derinitions are reduced and we get the conditional behavior.

ElectricMaxxx:
@dbu  i hopefully did what you whanted. I moved the enhancer sorting and loopging through the enhancers from the `DynamicRouter`  to a `ConditionalEnhancer`. For regular enhancers it works as before. For the enhancers that need parameters i created an interface which forces them to do a name comparison (am i responsible for a set of parameters) and to implement a setter for the mapping/values/parameters passed to the the enhancer. Those parametes do not need to be passed throug the service defintion anymore. The `ConditionalEnhancer` takes them all and pass them to the enhancers by asking them "is that yours" (`isName()`).

That comes with several advantages:
* no need to do the tagging of parameter depending enhancers in the extension bases of the configuration
* no need to pass the separated configuration parameters through the service
* chance to "normalize" the parameters based on a HTTP method in a generic way (all configuration can have a methods mapping now, no matter if that makes sense)
* the enhancers stay stable in their core behavior.
* `DynamicRouter` depends on an abstraction now, doesn't do it on its own

I hopefully get into the direction you whanted it, at least i hope this can be a starting point for a refactoring without reverting to much stuff.

PS: Some stuff (docblock, tests) is still missing.

dbu:
i tried to show what i imagined in #174

ElectricMaxxx:
solved in #174

PR #173, opened 2016-05-08
Fix randomly failing tests

WouterJ:


Tests/Routing/DynamicRouterTest.php
```diff
@@ -147,9 +147,12 @@ public function testMatchUrl()
         ;
 
         $expected = array('this' => 'that');
+        $test = $this;
         $this->enhancer->expects($this->once())
             ->method('enhance')
-            ->with($this->equalTo($routeDefaults), $this->equalTo($this->request))
+            ->with($this->equalTo($routeDefaults), $this->callback(function (Request $request) use ($test) {
```
ElectricMaxxx: There is some php closure fuckup in v5.3 here. Can't we pass the url directly?

WouterJ: fix it

ElectricMaxxx: uhh even better :-)

dbu:
thanks wouter!
what was it? timestamp issues with request objects?

WouterJ:
Yeah, but I'm not really sure why the timestamp changed (as why seem to used the same object instance twice).

PR #174, opened 2016-05-10
conditional enhancer

dbu:
alternative to #172 

@ElectricMaxxx this is what i had in mind. it would keep the changes minimal and have no impact for people not needing conditional enhancers.

### DIFF
--- a/Tests/Enhancer/ConditionalEnhancerTest.php
+++ b/Tests/Enhancer/ConditionalEnhancerTest.php
@@ -1,0 +1,92 @@
+<?php
+
+/*
+ * This file is part of the Symfony CMF package.
+ *
+ * (c) 2011-2015 Symfony CMF
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Cmf\Component\Routing\Tests\Mapper;
+
+use Symfony\Cmf\Component\Routing\Enhancer\ConditionalEnhancer;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Cmf\Component\Routing\Test\CmfUnitTestCase;
+
+class ConditionalEnhancerTest extends CmfUnitTestCase
+{
+    /**
+     * @var Request
+     */
+    private $request;
+
+    public function setUp()
+    {
+        $this->request = Request::create('/test');
+    }
+
+    public function testSecondMatch()
+    {
+        $defaults = array('foo' => 'bar');
+        $expected = array('matcher' => 'found');
+
+        $enhancer1 = $this->buildMock('\Symfony\Cmf\Component\Routing\Enhancer\RouteEnhancerInterface');
+        $enhancer1->expects($this->never())
+            ->method('enhance');
+        $matcher1 = $this->buildMock('\Symfony\Component\HttpFoundation\RequestMatcherInterface');
+        $matcher1->expects($this->once())
+            ->method('matches')
+            ->with($this->request)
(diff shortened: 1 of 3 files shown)


ElectricMaxxx:
That looks really nice, but i miss some points:

* How do you get that enhancerMap?
* Where would you pass the the mapping informormation, that tree of five current enhancers would need? Still through service defintion or the constructor
* The matcher just decides for the responsibility of the enhancer? So it would be aware of all "early outs" we currently have in the enhancers? Why can't the enhancer have a `match()` method?
* Where would you see the point of preparing mapping information like
```
[
    'methods' => ['put', 'post'],
   'value' => 'service:postAndPutAction'
]
```
into
```
[
    'service:postAndPutAction'
]
```
for POST/PUT request and nothing for all other requests? Caus the `['service:postAndPutAction']` is the only mapping the `FieldByClass` enhancer (with target=_controller)  is able to work with, to get the right controllerAction for the resolver.

dbu:
>   * How do you get that enhancerMap?

for me, the configuration is the job of the bundle, not of the library.
the library works with object graphs.

>   * Where would you pass the the mapping informormation, that tree of
>     five current enhancers would need? Still through service defintion
>     or the constructor

service definitions that are created from the configuration.

>   * The matcher just decides for the responsibility of the enhancer? So
>     it would be aware of all "early outs" we currently have in the
>     enhancers? Why can't the enhancer have a |match()| method?

i don't want to complicate the enhancers. the matcher is used to
determine if a specific enhancer is applicable for this request. the
enhancers should not need to know about that, they would be duplicating
the request matchers.

>   * Where would you see the point of preparing mapping information like
> 
> |[ 'methods' => ['put', 'post'], 'value' => 'service:postAndPutAction' ] |
> 
> into
> 
> |[ 'service:postAndPutAction' ] |
> 
> for POST/PUT request and nothing for all other requests? Caus the
> |['service:postAndPutAction']| is the only mapping the |FieldByClass|
> enhancer (with target=_controller) is able to work with, to get the
> right controllerAction for the resolver.

```php
$controllerEnhancer = new FieldByClassEnhancer(
    '_content',
    '_controller',
    ['Symfony\Cmf\Component\Routing\Tests\Enhancer\RouteObject' =>
'cmf_content.controller:postAndPutAction']
);
$putPostMatcher = new RequestMatcher(null, null, ['put', 'post']);

$conditionalEnhancer = new ConditionalEnhancer([
    ['matcher' => $putPostMatcher, 'enhancer' => $controllerEnhancer],
]);

```

because we add no further enhancer, this thing will only act on post/put
requests.

dbu:
fixed the bugs i had in this draft. i think its now ready to merge.

Tests/Enhancer/ConditionalEnhancerTest.php
```diff
@@ -0,0 +1,93 @@
...
+            ->method('enhance')
+            ->with($defaults, $this->request)
+            ->will($this->returnValue($expected));
+        $matcher2 = $this->buildMock('\Symfony\Component\HttpFoundation\RequestMatcherInterface');
+        $matcher2->expects($this->once())
+            ->method('matches')
+            ->with($this->request)
+            ->will($this->returnValue(true));
+
+
```
ElectricMaxxx: Can you please remove the second empty line to satisfy StyleCI and me.

dbu: fixed. i am now off until next wednesday

ElectricMaxxx:
thanks a lot for your effort to explain and solving the problem.

PR #175, opened 2016-05-15
implement adder and creator of enhancer map entries

ElectricMaxxx:
The conditional enhancer can not get its enhancer map with object when it is created by a symfony DI container. So i added a static creation method to create an entry and on adder method, to add by position. This position is important to, so we can force who comes first, as the first match will win.


ElectricMaxxx:
Can i keep that content fixture class as an empty fixture to work on in tests?

PR #182, opened 2016-12-03
created an empty upgrade file

ElectricMaxxx:


dbu:
do you want to already merge this? or is this to start adding things?

wouterj:
Empty is useless, either add upgrade notes or don't add an UPGRADE guide.

dbu:
looking at https://github.com/symfony-cmf/routing/compare/1.4.0...master i don't think we need an UPGRADE file. don't see any BC breaks. 

i am gathering input on #116. if we end up doing it, we should add the UPGRADE.md there

PR #187, opened 2017-01-19
Adder conditional enhancer

### LINKED ISSUE #176: Issues with conditional enhancer
To explain #175 and describe the problems with the conditional enhancer:
### Can't inject objects by DI container

Late, but when implementing the extension in routing bundle to use the conditional enhancer, i realized, that an enhancer map containing the objects of matcher and enhancer won't work as the DI container does not like any objects (dumping into xml problems). 
That was the reason to create #175 Currently i am not sure if i need the implemented `priority` - cause of the second problem
### Every class/type/<item> needs an own conditional enhancer

I would just start with an example:

```

        controllers_by_type:
            demo_alias: 
              -
                methods: ['put', 'post']
                value: test.controller:aliasAction
        controllers_by_class:
            Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Phpcr\RedirectRoute:
              -
                methods: ['put', 'post']
                value: cmf_routing.redirect_controller:redirectAction
            Symfony\Cmf\Bundle\RoutingBundle\Tests\Resources\Document\Content:
              - 
                methods: ['post', 'put']
                value: service:putAction
        templates_by_class:
            Symfony\Cmf\Bundle\RoutingBundle\Tests\Resources\Document\Content: TestBundle:Content:index.html.twig
```

without that method mapping that would leed to 4 different enhancers. 3 FieldByClass (controller_by_class, controller_by_type, controller_for_template_by_class) and 1 FieldMapPre.. 

i.e. PUT request:
I can't put the complete configuration into one conditional enhancer. The first request match would close the chain even when the enhancer except _type = demo_alias => _controller = test.controller:aliasAction.
So would need to create 3 different conditional ehnacers. One vor the target pair _type/_controller _content/_controller and one for null/_template. The _content/_controller pair can go of controller_for_template_by_class and controller_by_class can go in one - they have the same target/source pair - i thought. But again: The first request match would close the chain completely independent from the work of the enhancer.
So lets to a 4th conditional enhancer to put especially controller_by_class into one map. But lets expect a PUT request again on a route with content of class Symfony\Cmf\Bundle\RoutingBundle\Tests\Resources\Document\Content. The request matcher would close agein on the first entry the FieldByClass enhancer will see that the current content is of class Symfony\Cmf\Bundle\RoutingBundle\Doctrine\Phpcr\RedirectRoute and would do nothing. Chain closed :-(

So The only way to work with the conditional enhancer is by creating one for each type/class/<item> to just find the first request match. So i would create 5 different conditional enhancers for the current example. I don't know if that is a step forward. Is that the way we planned it @dbu? From the point of seperating concerns it is really good but from the performance view we instantiate two times classes as before.

ElectricMaxxx:
This will be the next try to enable the conditional enhancer. I will use it first to create an PR on the bundle trying to enable the conditional enhancer.

fix #176

### DIFF
--- a/src/Enhancer/ConditionalEnhancer.php
+++ b/src/Enhancer/ConditionalEnhancer.php
@@ -13,6 +13,7 @@ namespace Symfony\Cmf\Component\Routing\Enhancer;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestMatcher;
+use Symfony\Component\Routing\Matcher\RequestMatcherInterface;
 
 /**
  * This enhancer uses a HttpFoundation request matcher to decide which enhancer
@@ -61,7 +62,9 @@ class ConditionalEnhancer implements RouteEnhancerInterface
     public function enhance(array $defaults, Request $request)
     {
         foreach ($this->getEnhancerMap() as $pair) {
-            if ($pair[self::MAP_KEY_MATCHER]->matches($request)) {
+            if (($pair[self::MAP_KEY_MATCHER] instanceof RequestMatcherInterface
+                    && $pair[self::MAP_KEY_MATCHER]->matchRequest($request)
+                ) || $pair[self::MAP_KEY_MATCHER]->matches($request)) {
                 return $pair[self::MAP_KEY_ENHANCER]->enhance($defaults, $request);
             }
         }
--- a/src/Enhancer/ConditionalEnhancer.php
+++ b/src/Enhancer/ConditionalEnhancer.php
@@ -69,6 +69,11 @@ class ConditionalEnhancer implements RouteEnhancerInterface
         return $defaults;
     }
 
+    /**
+     * Creates a sorted list of enhancer map pairs. It is ordered by the position, which should be the key.
+     *
+     * @return array
+     */
     public function getEnhancerMap()
     {
         if (count($this->enhancerMap) === 0 || count($this->sortedEnhancerMap) > 0) {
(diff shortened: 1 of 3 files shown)


src/Enhancer/ConditionalEnhancer.php
```diff
@@ -48,12 +61,75 @@ public function __construct(array $enhancerMap)
...
+     * To create and add an enhancer map entry.
+     *
+     * @param string    $className FQCN of the enhancer.
+     * @param string    $source    Source value for the enhancer
+     * @param string    $target    Target value for the enhancer
+     * @param []|string $map       Map values for the enhancer
+     * @param []        $methods   List of methods the enhancer is restricted on.
+     * @param int       $position  To order the enhancer calls
+     */
+    public function createAndAddMapEntry($className, $source, $target, $map, $methods = null, $position = 0)
```
dbu: imo this should just be addMapEntry and expect the caller to do the request matcher and stuff

src/Enhancer/ConditionalEnhancer.php
```diff
@@ -34,11 +39,19 @@ class ConditionalEnhancer implements RouteEnhancerInterface
...
+     * @var []
+     */
+    private $sortedEnhancerMap = array();
+
+    /**
+     * @param array $enhancerMap Ordered list of 'matcher', 'enhancer' pairs.
+     * @param int   $position    Position to add in the list.
      */
-    public function __construct(array $enhancerMap)
+    public function __construct(array $enhancerMap = array(), $position = 0)
```
dbu: are you sure we want to make the initial map optional? even if we keep it required, users can still call this with an explicit empty array if they don't know at that point...

on the other hand, why would we need to dynamically add more things to the enhancer after it has been created? i think all of this should be determined at bootstrap. ah, but this map is only for one position...

src/Enhancer/ConditionalEnhancer.php
```diff
@@ -48,12 +61,75 @@ public function __construct(array $enhancerMap)
      */
     public function enhance(array $defaults, Request $request)
     {
-        foreach ($this->enhancerMap as $pair) {
-            if ($pair['matcher']->matches($request)) {
-                return $pair['enhancer']->enhance($defaults, $request);
+        foreach ($this->getEnhancerMap() as $pair) {
+            if (($pair[self::MAP_KEY_MATCHER] instanceof RequestMatcherInterface
```
dbu: i notice that i got confused (again) by HttpFoundation and Routing both having a RequestMatcherInterface but with different method name... we should decide which of the two makes sense (i think the routing one) and force to only use that. no use in mixing the two kind of request matcher here.

dbu:
you will need to revert cc6325e67ae4b7c90891484b588b6da996c03625 to get this branch working again. see #188

ElectricMaxxx:
we'll see what happens

PR #188, opened 2017-01-20
remove conditional enhancer that was never released

dbu:
i propose to remove the conditional enhancer for the 2.0 release and add it back in 2.1 when we know how exactly it should work.


wouterj:
👍

PR #189, opened 2017-01-22
enable more styleci checks

dbu:


dbu:
this leads to #190 - do we agree with these style rules? i can then apply that to all repos...

Issue #194, opened 2017-05-08
Candidates::determineLocale regex matches wrongly


uwej711:
Hi all,

I'm currently upgrading one of our projects and ran into an issue with 
https://github.com/symfony-cmf/routing/blob/master/src/Candidates/Candidates.php#L111

The regular expression there matches my url "/buchen" if I have "en" in the list of locales. I guess it should be 
```
if (preg_match('#^/('.implode('|', $this->locales).')(/|$)#', $url, $matches)) {
```
instead.

uwej711:
I can do a PR against 1.4

dbu:
ouch, thats quite bad! yes please do a pr and i will merge it to master if
its still present there

Am 08.05.2017 1:58 nachm. schrieb "Uwe Jäger" <notifications@github.com>:

uwej711:
I would be happy if that would go in a 1.4.1 version as well ... spares me maintaining a fork.

dbu:
fixed in #195

PR #195, opened 2017-05-08
fix determine locales regular expression

uwej711:
See #194 
I would have made a PR against 1.4 but there is no branch 1.4


dbu:
i created the 1.4 branch from the 1.4 tag and cherry-picked this commit. php 7 build now fails because changes on travis-ci but i will tag a 1.4.1. i then merged 1.4 to master so we have the changes in master as well.

dbu:
thank you very much for the fix uwe! i tagged 1.4.1 and 2.0.1 now

PR #196, opened 2017-06-19
Refactored Travis configuration file

damienflament:
What's about the [email notifications](https://github.com/symfony-cmf/routing/pull/196/files?diff=split#diff-354f30a63fb0907d4ad57269548329e3R44) ?

Only the IRC notifications are configured for the _Routing Auto_ component/bundle repository. Is it an oversight ? Or is this email no more used ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: php
-
+dist: trusty
 php:
   - 7.0
   - hhvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
 matrix:
   include:
     - php: 7.1
-      env: SYMFONY_VERSION=3.3.* DEPS=dev
+      env: SYMFONY_VERSION=3.3.* STABILITY=dev
     - php: 5.6
       env: SYMFONY_VERSION=2.8.* COMPOSER_FLAGS="--prefer-lowest"
     - php: 7.1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: php
-
+dist: trusty
```
dbu: please instead remove hhvm from the `php` list and add another entry in matrix: include: that has php: hhvm and dist: trusty lines. that way we only change the dist for hhvm.

damienflament: I didn't know that the `dist` key can be used within matrices inclusion. I thought only `env` and `php` was available.

So we need to do the same within the other repositories. See those PR which enable Trusty for all jobs:
symfony-cmf/routing-auto/pull/92
symfony-cmf/routing-auto-bundle/pull/203

We should think about a way to share a  common Travis configuration file for all repositories. This is becoming tedious to maintain its configuration for just 4 repositories. I can't image the headache after updating it for all Symfony CMF repositories.

It explains why those files are so different. Not only in there formatting. But also concerning the PHPUnit distribution type, its version,.

damienflament: I should mention that actually, all the build on HHVM may fail due to its unavailability on Ubuntu Precise. So, all Travis configuration files need to be updated.

damienflament:
I'm closing this PR as I cannot properly refactor the Travis configuration files this way.

I will work on a POC on a fork of the _symfony-cmf/testing_ component in order to use a common Travis configuration file (maybe two, one for components and one for bundles). My idea is to override this common  configuration within the individual configuration file of the repository. A simple command will merge them.

The [travis-ci/travis-yaml](https://github.com/travis-ci/travis-yaml) library might be interesting. But as it is written in Ruby, I prefer to start experimentation with [symfony/yaml](https://github.com/symfony/yaml).

I close symfony-cmf/routing-auto-bundle#205 also.

dbu:
sonata has built a dev-kit to manage these files across their many repositories: https://github.com/sonata-project/dev-kit

that could also be a possible approach to centrally manage these meta data things. (i assume we would have to fork their repo, rather than trying to super-generalize it, but have not looked closely into it)

damienflament:
I will look at it. Thanks !

PR #200, opened 2017-09-26
PHP 7.2 compatibility: check if sortedRouters is null before trying to count it

maff:
Starting from PHP 7.2, `count(null)` yields a warning (see https://3v4l.org/hbmpC).

Alternatively `sortedRouters` could be initialized as empty array and trigger a sort if count of `sortedRouters` differs from `routers` , but I don't know the exact background of this logic.

### DIFF
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -47,7 +47,7 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
     /**
      * @var RouterInterface[] Array of routers, sorted by priority
      */
-    private $sortedRouters;
+    private $sortedRouters = [];
 
     /**
      * @var RouteCollection


dbu:
thank you for checking this and providing a fix! lets initialize in the property declaration with an empty array, its also what the add() method does to reset the sortedRouter field. can you please adjust the pull request accordingly?

maff:
Ah right, didn't see that in the `add()`method. PR is updated.

dbu:
thanks a lot. i will tag 2.0.2 with this today.

dbu:
i now tagged.

i wanted to run the tests with php 7.2 too so we can be sure its alright, but unfortunately there is a failure in the dependencies that i have troubles figuring out. https://github.com/phpcr/phpcr-utils/pull/174

PR #202, opened 2017-09-27
test with php 7.2

dbu:


dbu:
this is already done in master.

PR #203, opened 2017-09-28
Make ChainRouter work with PHP 7.2

dkreuer:
Currently the chain router fails in symfony context with 
```
In ChainRouter.php line 101:
                                                                                       
  [Symfony\Component\Debug\Exception\ContextErrorException]                            
  Warning: count(): Parameter must be an array or an object that implements Countable

```

This is because the sortedRouters property is not initialized.

The change proposes to take the not initialized property into account while fetching all routes.

### DIFF
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -98,7 +98,7 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
      */
     public function all()
     {
-        if (0 === count($this->sortedRouters)) {
+        if (null === $this->sortedRouters || 0 === count($this->sortedRouters)) {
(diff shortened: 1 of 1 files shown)


dbu:
this has been fixed in #200 already. i then wanted to actually test that everything works with php 7.2 before tagging a release, but got sidetracked by things not working in one of the phpcr dependencies. (https://github.com/phpcr/phpcr-utils/pull/174 if you are curious). as that problem is not trivial to figure out, and we already claim to be compatible with php 7.2, i will just tag 2.0.2 now as its definitely an improvement.

dbu:
2.0.2 is now tagged

dkreuer:
Thank you very much.

PR #206, opened 2017-11-16
DevKit updates

ElectricMaxxx:
some more clean ups.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,8 @@
 Contributing
 ------------
 
-Symfony2 CMF is an open source, community-driven project. We follow the same
-guidelines as core Symfony2. If you'd like to contribute, please read the
+Symfony CMF is an open source, community-driven project. We follow the same
+guidelines as core Symfony. If you'd like to contribute, please read the
 [Contributing Code][1] part of the documentation. If you're submitting a pull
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -1,8 +1,8 @@
 Contributing
 ------------
 
-Symfony CMF is an open source, community-driven project. We follow the same
-guidelines as core Symfony. If you'd like to contribute, please read the
+Symfony2 CMF is an open source, community-driven project. We follow the same
+guidelines as core Symfony2. If you'd like to contribute, please read the
```
dbu: there is some regression here

ElectricMaxxx:
Indeed that seems to be a problem, i will look if dev-kit uses a fresh master always or an old one.

ElectricMaxxx:
I think it was me when fetching the new master-dev-kit branch from next try. I had the old one. Shall we subfix them with a kind of a hash which should be stable for the current run? to force me to fetch a new branch every time?

CONTRIBUTING.md
ElectricMaxxx: that is strange

ElectricMaxxx:
Travis is green, but does not show it here. :-(

dbu:
please update the php version in composer.json as well, so it matches what we do on travis-ci

PR #211, opened 2017-11-18
DevKit updates

ElectricMaxxx:
Just to have a valid base if it works with symfony 4.0.


ElectricMaxxx:
It will definitly fail. but i will merge to serve a master to fix :-)

PR #215, opened 2017-12-07
avoid error with non-array parameters

dbu:
The UrlGeneratorInterface::generate is not very specific about $parameters and allows non-arrays.
I don't want to get more fancy than this, its only about a very edge-case in error reporting in the end.

| Q             | A
| ------------- | ---
| Branch?       | 1.4
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -


dbu:
i was confused, this is a super old branch. going to version 2 now...

Issue #218, opened 2018-03-15
Uninitialised requestContext leads to "Call to a member function setParameter() on null -- in LocaleListener.php"


tapetersen:
### Environment

dev
#### Symfony packages

```
symfony/asset                v4.0.6             v4.0.6             Symfony Asset Component
symfony/browser-kit          v4.0.6             v4.0.6             Symfony BrowserKit Component
symfony/cache                v4.0.6             v4.0.6             Symfony Cache component with PSR-6, PSR-16, and tags
symfony/config               v4.0.6             v4.0.6             Symfony Config Component
symfony/console              v4.0.6             v4.0.6             Symfony Console Component
symfony/css-selector         v4.0.6             v4.0.6             Symfony CssSelector Component
symfony/debug                v4.0.6             v4.0.6             Symfony Debug Component
symfony/debug-bundle         v4.0.6             v4.0.6             Symfony DebugBundle
symfony/debug-pack           v1.0.4             v1.0.4             A debug pack for Symfony projects
symfony/dependency-injection v4.0.6             v4.0.6             Symfony DependencyInjection Component
symfony/doctrine-bridge      v4.0.6             v4.0.6             Symfony Doctrine Bridge
symfony/dom-crawler          v4.0.6             v4.0.6             Symfony DomCrawler Component
symfony/dotenv               v4.0.6             v4.0.6             Registers environment variables from a .env file
symfony/event-dispatcher     v4.0.6             v4.0.6             Symfony EventDispatcher Component
symfony/expression-language  v4.0.6             v4.0.6             Symfony ExpressionLanguage Component
symfony/filesystem           v4.0.6             v4.0.6             Symfony Filesystem Component
symfony/finder               v4.0.6             v4.0.6             Symfony Finder Component
symfony/flex                 v1.0.71            v1.0.71           
symfony/form                 v4.0.6             v4.0.6             Symfony Form Component
symfony/framework-bundle     v4.0.6             v4.0.6             Symfony FrameworkBundle
symfony/http-foundation      v4.0.6             v4.0.6             Symfony HttpFoundation Component
symfony/http-kernel          v4.0.6             v4.0.6             Symfony HttpKernel Component
symfony/inflector            v4.0.6             v4.0.6             Symfony Inflector Component
symfony/intl                 v4.0.6             v4.0.6             A PHP replacement layer for the C intl extension that includes additional data from the ICU library.
[...]

PR #219, opened 2018-03-15
Initialize context in constructor of chainrouter.

tapetersen:
Initialize the requestcontext in the contstructor of the chainrouter to avoid getContext() to return null

| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #218 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -64,7 +64,8 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
      */
     public function __construct(LoggerInterface $logger = null)
     {
-        $this->logger = $logger;
+        $this->logger = $logger;      
+        $this->context = new RequestContext();
     }
 
     /**


dbu:
hi, thanks for the pull request! and thanks for the good explanation in #218 - on seeing the PR i was unsure if this is the right thing to do, but #218 convinced me its the right thing (mainly that the symfony router creates an empty context).

there is one test failing, and i think its because of https://github.com/tapetersen/Routing/blob/b1356fdc09eda890ecf9068905bbb8722e061daa/src/ChainRouter.php#L108 - setContext calls `all()` which sets the context on all routers if it is not null. that means that we will now set the empty context on all routers and then immediately set the new context. the quick fix would be to first set the context to null before calling `all` to avoid `all` setting the context. alternatively, we could extract updating the context into a private method and call that from all and from setContext (because sorting the routers is not necessary for setContext).

can you have a look which way we can best do this?

src/ChainRouter.php
```diff
@@ -64,7 +64,8 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
      */
     public function __construct(LoggerInterface $logger = null)
     {
-        $this->logger = $logger;
+        $this->logger = $logger;
```
dbu: styleci is unhappy because there is trailing whitespace added. the other failures are not related, i will fix those on master.

dbu: fixed in master. if you remove the trailing whitespace and rebase, it should get green here.

tapetersen: I can fix the whitespace errors, regarding the check for null I would suggest just removing it as initializing it in the constructor should take care of the problem of needing it uninitialized.

Another thought (havn't checked) is if the change is needed in DynamicRouter as  well.

tapetersen: Hmm I would tend to think the correct solution is to always set the context in all() regardless if it's null or not but I have a hard time getting it to play nicely with the test. Of course the loop could check to see if the context is already the same before trying to set it but that doesn't really seem like the ideal solution either.

I tried to avoid using once() and just comparing the contexts afterwards but the mocks don't keep the values.

I unfortunately don't really have much more time to spend on this (having solved my real problem in another way) and of course you could argue the bug is in localeAwareListener and that there is nothing guaranteeing ->getContext actually returns a non-null object but that is kind of problematic as the standard Router de facto does have that behavior.

(My main obstacle is inexperience with phpunit mock-object's and the structuring of the tests)

I may have time to look at this in the future but not this week.

Thanks anyway for writing good code and keeping up.

dbu:
> Another thought (havn't checked) is if the change is needed in DynamicRouter as well.

the DynamicRouter requires the context in the constructor, it can't be created without context.

thanks for bringing this up. i will try to find a solution to only set the context once.

PR #234, opened 2019-02-12
test current symfony versions

dbu:
we claim to work with those versions, lets see if that is true.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,10 @@ env:
 matrix:
   include:
     - php: 7.2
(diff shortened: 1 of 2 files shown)


dbu:
@ElectricMaxxx so there seems to be some regression with symfony 4.1 according to the test matrix. maybe you should discuss the ticket you created with this PR, as it shows since when the problem appears. looking at https://github.com/symfony/symfony/blob/master/CHANGELOG-4.1.md there are a bunch of changes to routing matching (the thing about compiling a regular expression could be the culprit, or one of the things about trailing slash possibly)?

Issue #236, opened 2019-06-05
Wrong type hint for ChainRouter::generate


korsarNek:
symfony-cmf/routing 2.1.0

## Problem

The ChainRouter::generate function uses inheritdoc for the base PHPDoc, but UrlGeneratorInterace::generate says that name must be a string. While with ChainRouter::generate it could be anything.

That is a problem if tools are used that check the PHPDoc types too to validate code, like PHPStan.

## Steps to reproduce

I don't have an example project right now, but having a project with PHPStan enabled and using the ChainRouter::generate function from anywhere should show the same error.

## Expected results

The ChainRouter::generate function should change the $name parameter type to mixed.

## Actual results

PHPStan shows this error:
```
Parameter #1 $name of method eZ\Publish\Core\MVC\Symfony\Routing\ChainRouter::generate() expects string, eZ\Publish\API\Repository\Values\Content\Location given.
```
The eZ\Publish ChainRouter itself has no PHPDoc, it is taken from the UrlGeneratorInterface.

korsarNek:
Sorry, wrong alarm. It was in another package.

dbu:
please do a pull request to change the UrlGeneratorInterface phpdoc (and try it out in your project to be sure it works correctly). i am not sure how well phpstan solves such disagreements between phpdoc for the same method, but i guess it should work.

korsarNek:
I've created the pull request, unfortunately PHPStan can not handle merging of PHPDocs, so I copied the PHPDoc from the UrlGeneratorInterface and modified it a bit.

dbu:
i fixed this in #252

PR #237, opened 2019-06-19
Fixing https://github.com/symfony-cmf/Routing/issues/236

korsarNek:
The ChainRouter::generate function has a wider type for the name parameter than
the UrlGeneratorInterface.

Some analysis tools like PHPStan check the PHPDoc to make sure the correct
types are used.

With this change the PHPDoc for ChainRouter::generate reflects the widened type.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 236
| License       | MIT

### DIFF
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -14,7 +14,9 @@ namespace Symfony\Cmf\Component\Routing;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface;
+use Symfony\Component\Routing\Exception\InvalidParameterException;
 use Symfony\Component\Routing\Exception\MethodNotAllowedException;
+use Symfony\Component\Routing\Exception\MissingMandatoryParametersException;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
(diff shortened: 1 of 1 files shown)


src/ChainRouter.php
```diff
@@ -211,10 +213,34 @@ private function doMatch($pathinfo, Request $request = null)
...
+     * ABSOLUTE_URL with the https scheme and the current host. This makes sure the generated URL matches
+     * the route in any case.
+     *
+     * If there is no route with the given name, the generator must throw the RouteNotFoundException.
+     *
+     * The special parameter _fragment will be used as the document fragment suffixed to the final URL.
+     *
+     * @param mixed $name       the name of the route or something identifying a route that can be used by any of
+     *                          the chained routers
+     * @param mixed $parameters An array of parameters
```
dbu: shouldn't this be `@param array $parameters`?

src/ChainRouter.php
```diff
@@ -211,10 +213,34 @@ private function doMatch($pathinfo, Request $request = null)
...
+     * When the passed reference type cannot be generated for the route because it requires a different
+     * host or scheme than the current one, the method will return a more comprehensive reference
+     * that includes the required params. For example, when you call this method with $referenceType = ABSOLUTE_PATH
+     * but the route requires the https scheme whereas the current scheme is http, it will instead return an
+     * ABSOLUTE_URL with the https scheme and the current host. This makes sure the generated URL matches
+     * the route in any case.
+     *
+     * If there is no route with the given name, the generator must throw the RouteNotFoundException.
+     *
+     * The special parameter _fragment will be used as the document fragment suffixed to the final URL.
```
dbu: could you keep using the `{@inheritdoc}` instead of this explanations but keep the new parameter annotations below? or is phpstan getting confused as soon as it sees an `@inheritdoc`?

src/ChainRouter.php
korsarNek: makes sense, I just copied it from the UrlGeneratorInterface definition. Should I also change it there?

korsarNek: Ahh, that's actually in another repository. I'll just change it here.

dbu: can you try if `@inheritdoc` works to allow you to only specify the argument that actually is different?

src/ChainRouter.php
korsarNek: Unfortunately, PHPStan gets confused if I want to use inheritdoc and anything else at the same time.

src/ChainRouter.php
```diff
@@ -211,10 +213,34 @@ private function doMatch($pathinfo, Request $request = null)
...
+     * path or host. Extra params are added as query string to the URL.
+     *
+     * When the passed reference type cannot be generated for the route because it requires a different
+     * host or scheme than the current one, the method will return a more comprehensive reference
+     * that includes the required params. For example, when you call this method with $referenceType = ABSOLUTE_PATH
+     * but the route requires the https scheme whereas the current scheme is http, it will instead return an
+     * ABSOLUTE_URL with the https scheme and the current host. This makes sure the generated URL matches
+     * the route in any case.
+     *
+     * If there is no route with the given name, the generator must throw the RouteNotFoundException.
```
dbu: lets at least remove the whole text block and use an `@see` annotation pointing to the parent class method instead. i don't want to duplicate this text and it should not matter for phpstan. can you do that (and check if phpstan is still happy?)

src/ChainRouter.php
dbu: hm, unfortunate. okay.

dbu:
is this fixed with #250 ?

acrobat:
@dbu I think this can be closed now that the symfony 5 support pr's are merged!

dbu:
symfony 5 support means that we deprecated to pass non-strings to the route generator. thanks to type widening, the chain router still can handle them for BC, but it is no longer recommended. i did #252 to clarify things in the phpdoc.

PR #240, opened 2019-11-23
[WIP] Add support for symfony 5 #SymfonyHackday

vincentmoulene:
| Q             | A
| ------------- | ---
| Branch?       | "master" 
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

#SymfonyHackday
Pull request made for the support of Symfony 5. 
I am a beginner in open source, so let me know if everything is good.

I saw the tests failed with another component `symfony-cmf/testing` but I don't know how replace it in the `composer.json` file to test with another branch.
Let me know and I will update this PR.

Thanks

PR #245, opened 2020-01-06
[WIP] Add bc layers for symfony 5 support

acrobat:
| Q             | A
| ------------- | ---
| Branch?       | "master"
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 
| License       | MIT
| Doc PR        | TODO

TODO:

- [ ] Update docs for generate name parameter deprecation
- [x] Check if the tests cover this code correctly and verify solution

@dbu this is a first try to work around the generate method signature in compatibility (+ some extra fixes for sf5 support). What do you think?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "symfony-cmf/testing": "^3@dev"
     },
     "suggest": {
-        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^3.4 || ^4.3)"
+        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^3.4 || ^4.3 || ^5.0‡ )"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "symfony-cmf/testing": "^3@dev"
     },
     "suggest": {
-        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^3.4 || ^4.3 || ^5.0‡ )"
+        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^3.4 || ^4.3 || ^5.0)"
     },
     "autoload": {
         "psr-4": {
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -30,7 +30,7 @@ use Symfony\Component\Routing\RouterInterface;
  * @author Henrik Bjornskov <henrik@bjrnskov.dk>
  * @author Magnus Nordlander <magnus@e-butik.se>
  */
-class ChainRouter implements ChainRouterInterface, WarmableInterface
+class ChainRouter extends ChainRouterGenerateBcLayer implements ChainRouterInterface, WarmableInterface
 {
     /**
      * @var RequestContext|null
@@ -211,49 +211,6 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
     }
 
     /**
-     * {@inheritdoc}
-     *
-     * Loops through all registered routers and returns a router if one is found.
-     * It will always return the first route generated.
-     */
-    public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
-    {
-        $debug = [];
-
-        foreach ($this->all() as $router) {
-            // if $router does not announce it is capable of handling
-            // non-string routes and $name is not a string, continue
-            if ($name && !is_string($name) && !$router instanceof VersatileGeneratorInterface) {
-                continue;
-            }
-
-            // If $router is versatile and doesn't support this route name, continue
-            if ($router instanceof VersatileGeneratorInterface && !$router->supports($name)) {
-                continue;
-            }
-
-            try {
-                return $router->generate($name, $parameters, $absolute);
-            } catch (RouteNotFoundException $e) {
-                $hint = $this->getErrorMessage($name, $router, $parameters);
-                $debug[] = $hint;
-                if ($this->logger) {
-                    $this->logger->debug('Router '.get_class($router)." was unable to generate route. Reason: '$hint': ".$e->getMessage());
-                }
-            }
-        }
-
-        if ($debug) {
-            $debug = array_unique($debug);
-            $info = implode(', ', $debug);
-        } else {
-            $info = $this->getErrorMessage($name);
-        }
-
-        throw new RouteNotFoundException(sprintf('None of the chained routers were able to generate route: %s', $info));
-    }
-
-    /**
      * Rebuild the request object from a URL with the help of the RequestContext.
      *
      * If the request context is not set, this returns the request object built from $pathinfo.
@@ -284,26 +241,6 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
         $uri = $context->getScheme().'://'.$host.$uri.'?'.$context->getQueryString();
 
         return Request::create($uri, $context->getMethod(), $context->getParameters(), [], [], $server);
-    }
-
-    private function getErrorMessage($name, $router = null, $parameters = null)
-    {
-        if ($router instanceof VersatileGeneratorInterface) {
-            // the $parameters are not forced to be array, but versatile generator does typehint it
-            if (!is_array($parameters)) {
-                $parameters = [];
-            }
-            $displayName = $router->getRouteDebugMessage($name, $parameters);
-        } elseif (is_object($name)) {
-            $displayName = method_exists($name, '__toString')
-                ? (string) $name
-                : get_class($name)
-            ;
-        } else {
-            $displayName = (string) $name;
-        }
-
-        return "Route '$displayName' not found";
     }
 
     /**
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -29,7 +29,7 @@ use Symfony\Component\Routing\RouterInterface;
  * @author Henrik Bjornskov <henrik@bjrnskov.dk>
  * @author Magnus Nordlander <magnus@e-butik.se>
  */
-class ChainRouter extends ChainRouterGenerateBcLayer implements ChainRouterInterface, WarmableInterface
+class ChainRouter extends ChainRouterBcLayer implements ChainRouterInterface, WarmableInterface
 {
     /**
      * @var RequestContext|null
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -16,7 +16,6 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface;
 use Symfony\Component\Routing\Exception\MethodNotAllowedException;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
-use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Routing\Matcher\RequestMatcherInterface;
 use Symfony\Component\Routing\RequestContext;
--- a/src/ChainRouterBcLayer.php
+++ b/src/ChainRouterBcLayer.php
@@ -14,12 +14,9 @@ namespace Symfony\Cmf\Component\Routing;
 use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
-$refl = new \ReflectionClass(UrlGeneratorInterface::class);
-$generateMethod = $refl->getMethod('generate');
-$methodParameters = $generateMethod->getParameters();
-/** @var \ReflectionParameter $nameParameter */
-$nameParameter = array_shift($methodParameters);
-if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->getType()) {
+$urlGeneratorParameters = (new \ReflectionClass(UrlGeneratorInterface::class))->getMethod('generate')->getParameters();
+$urlGeneratorNameParameter = array_shift($urlGeneratorParameters);
+if ($urlGeneratorNameParameter && $urlGeneratorNameParameter->hasType() && 'string' === $urlGeneratorNameParameter->getType()) {
     /**
      * @internal
      */
@@ -75,7 +72,7 @@ abstract class ChainRouterBaseBcLayer
             }
 
             $routeName = $name;
-            if (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
+            if (RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name && array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
                 $routeName = $parameters[RouteObjectInterface::ROUTE_OBJECT];
             }
 
--- a/src/ChainRouterBcLayer.php
+++ b/src/ChainRouterBcLayer.php
@@ -23,7 +23,7 @@ if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->
     /**
      * @internal
      */
-    class ChainRouterGenerateBcLayer extends ChainRouterBaseBcLayer
+    class ChainRouterBcLayer extends ChainRouterBaseBcLayer
     {
         /**
          * Loops through all registered routers and returns a router if one is found.
@@ -38,7 +38,7 @@ if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->
     /**
      * @internal
      */
-    class ChainRouterGenerateBcLayer extends ChainRouterBaseBcLayer
+    class ChainRouterBcLayer extends ChainRouterBaseBcLayer
     {
         /**
          * Loops through all registered routers and returns a router if one is found.
--- a/src/ChainRouterBcLayer.php
+++ b/src/ChainRouterBcLayer.php
@@ -47,9 +47,9 @@ if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->
         public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
         {
             if (!is_string($name)) {
-                @trigger_error('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "_cmf_route" parameter in the parameters array.', E_USER_DEPRECATED);
+                @trigger_error(sprintf('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "%s" parameter in the parameters array.', RouteObjectInterface::ROUTE_OBJECT), E_USER_DEPRECATED);
 
-                $parameters['_cmf_route'] = $name;
+                $parameters[RouteObjectInterface::ROUTE_OBJECT] = $name;
                 $name = '';
             }
 
@@ -70,13 +70,13 @@ abstract class ChainRouterBaseBcLayer
         foreach ($this->all() as $router) {
             // if $router does not announce it is capable of handling
             // non-string routes and $name is not a string, continue
-            if (array_key_exists('_cmf_route', $parameters) && is_object($parameters['_cmf_route']) && !$router instanceof VersatileGeneratorInterface) {
+            if (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT]) && !$router instanceof VersatileGeneratorInterface) {
                 continue;
             }
 
             $routeName = $name;
-            if (array_key_exists('_cmf_route', $parameters) && is_object($parameters['_cmf_route'])) {
-                $routeName = $parameters['_cmf_route'];
+            if (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
+                $routeName = $parameters[RouteObjectInterface::ROUTE_OBJECT];
             }
 
             // If $router is versatile and doesn't support this route name, continue
--- a/src/ChainRouterBcLayer.php
+++ b/src/ChainRouterBcLayer.php
@@ -47,10 +47,10 @@ if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->
         public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
         {
             if (!is_string($name)) {
-                @trigger_error(sprintf('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "%s" parameter in the parameters array.', RouteObjectInterface::ROUTE_OBJECT), E_USER_DEPRECATED);
+                @trigger_error(sprintf('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass the `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME` constant as the route name and the object as "%s" parameter in the parameters array.', RouteObjectInterface::ROUTE_OBJECT), E_USER_DEPRECATED);
 
                 $parameters[RouteObjectInterface::ROUTE_OBJECT] = $name;
-                $name = '';
+                $name = RouteObjectInterface::OBJECT_BASED_ROUTE_NAME;
             }
 
             return $this->doGenerate($name, $parameters, $absolute);
--- a/src/ChainRouterGenerateBcLayer.php
+++ b/src/ChainRouterGenerateBcLayer.php
@@ -23,7 +23,7 @@ if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->
     /**
      * @internal
      */
-    class ChainRouterGenerateBcLayer
+    class ChainRouterGenerateBcLayer extends ChainRouterBaseBcLayer
     {
         /**
          * Loops through all registered routers and returns a router if one is found.
@@ -31,66 +31,14 @@ if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->
          */
         public function generate(string $name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
         {
-            $debug = [];
-
-            foreach ($this->all() as $router) {
-                // if $router does not announce it is capable of handling
-                // non-string routes and $name is not a string, continue
-                if (array_key_exists('_cmf_route', $parameters) && is_object($parameters['_cmf_route']) && !$router instanceof VersatileGeneratorInterface) {
-                    continue;
-                }
-
-                // If $router is versatile and doesn't support this route name, continue
-                if ($router instanceof VersatileGeneratorInterface && !$router->supports($name)) {
-                    continue;
-                }
-
-                try {
-                    return $router->generate($name, $parameters, $absolute);
-                } catch (RouteNotFoundException $e) {
-                    $hint = $this->getErrorMessage($name, $router, $parameters);
-                    $debug[] = $hint;
-                    if ($this->logger) {
-                        $this->logger->debug('Router '.get_class($router)." was unable to generate route. Reason: '$hint': ".$e->getMessage());
-                    }
-                }
-            }
-
-            if ($debug) {
-                $debug = array_unique($debug);
-                $info = implode(', ', $debug);
-            } else {
-                $info = $this->getErrorMessage($name);
-            }
-
-            throw new RouteNotFoundException(sprintf('None of the chained routers were able to generate route: %s', $info));
-        }
-
-        private function getErrorMessage($name, $router = null, $parameters = null)
-        {
-            if ($router instanceof VersatileGeneratorInterface) {
-                // the $parameters are not forced to be array, but versatile generator does typehint it
-                if (!is_array($parameters)) {
-                    $parameters = [];
-                }
-                $displayName = $router->getRouteDebugMessage($name, $parameters);
-            } elseif (is_object($name)) {
-                $displayName = method_exists($name, '__toString')
-                    ? (string) $name
-                    : get_class($name)
-                ;
-            } else {
-                $displayName = (string) $name;
-            }
-
-            return "Route '$displayName' not found";
+            return $this->doGenerate($name, $parameters, $absolute);
         }
     }
 } else {
     /**
      * @internal
      */
-    class ChainRouterGenerateBcLayer
+    class ChainRouterGenerateBcLayer extends ChainRouterBaseBcLayer
     {
         /**
          * Loops through all registered routers and returns a router if one is found.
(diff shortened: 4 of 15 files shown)


acrobat:
Hmm @dbu I'm actually not sure what we should do when a route object is passed as the name parameter 😕

tests/Unit/Routing/ChainRouterTest.php
```diff
@@ -633,9 +636,16 @@ public function testGenerateObjectNotFound()
 
     /**
      * A versatile router will generate the debug message.
+     *
+     * @group legacy
+     * @expectedDeprecation Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "_cmf_route" parameter in the parameters array.
      */
     public function testGenerateObjectNotFoundVersatile()
     {
+        if (!class_exists(Symfony\Component\Routing\Loader\ObjectRouteLoader::class)) {
```
dbu: lets use `use` for these

src/DynamicRouterGenerateBcLayer.php
```diff
@@ -0,0 +1,82 @@
...
+$methodParameters = $generateMethod->getParameters();
+/** @var \ReflectionParameter $nameParameter */
+$nameParameter = array_shift($methodParameters);
+if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->getType()) {
+    /**
+     * @internal
+     */
+    class DynamicRouterGenerateBcLayer
+    {
+        public function generate(string $name, $parameters = [], $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
```
dbu: as the only difference is the method signature, i would prefer to have a base class with a `doGenerate` method with the permissive signature, and then the if/else for the actual DynamicRouter class that has the right signature and calls doGenerate. that way there won't be code duplication

src/DynamicRouter.php
```diff
@@ -223,8 +192,17 @@ public function match($pathinfo)
 
         $request = Request::create($pathinfo);
         if ($this->eventDispatcher) {
+            $eventName = Events::PRE_DYNAMIC_MATCH;
             $event = new RouterMatchEvent();
-            $this->eventDispatcher->dispatch(Events::PRE_DYNAMIC_MATCH, $event);
+
+            // LegacyEventDispatcherProxy exists in Symfony >= 4.3
+            if (class_exists(LegacyEventDispatcherProxy::class)) {
```
dbu: lets add a private dispatchEvent method to the router class to extract this logic to a separate place.

for a moment i was wondering if we need the old way when we are with symfony 5 routing, but probably we should, as we can't prevent routing 5 from being installed together with event dispatcher 4

src/ChainRouterGenerateBcLayer.php
```diff
@@ -0,0 +1,160 @@
...
+    {
+        /**
+         * Loops through all registered routers and returns a router if one is found.
+         * It will always return the first route generated.
+         */
+        public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
+        {
+            $debug = [];
+
+            if ($name && !is_string($name)) {
```
dbu: do we need the first check? `is_string(null)` works and correctly returns false.

src/ChainRouterGenerateBcLayer.php
```diff
@@ -0,0 +1,160 @@
...
+
+            foreach ($this->all() as $router) {
+                // if $router does not announce it is capable of handling
+                // non-string routes and $name is not a string, continue
+                if (array_key_exists('_cmf_route', $parameters) && is_object($parameters['_cmf_route']) && !$router instanceof VersatileGeneratorInterface) {
+                    continue;
+                }
+
+                // If $router is versatile and doesn't support this route name, continue
+                if ($router instanceof VersatileGeneratorInterface && !$router->supports($name)) {
```
dbu: this would be incorrect, we should be using $parameters['_cmf_route']

src/ChainRouterGenerateBcLayer.php
```diff
@@ -0,0 +1,160 @@
...
+                @trigger_error('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "_cmf_route" parameter in the parameters array.', E_USER_DEPRECATED);
+            }
+
+            foreach ($this->all() as $router) {
+                // if $router does not announce it is capable of handling
+                // non-string routes and $name is not a string, continue
+                if ((($name && !is_string($name)) || (array_key_exists('_cmf_route', $parameters) && is_object($parameters['_cmf_route']))) && !$router instanceof VersatileGeneratorInterface) {
+                    continue;
+                }
+
```
dbu: for BC reasons, we should replace empty string $name with the value of _cmf_route here. otherwise the supports call only works with the deprecated mode, and old generators won't work.

acrobat: I've added a different bc layer on line 52, so we already use the new way. Old code will still work as the generators will still check the name for being an object

acrobat:
@dbu thanks for the review! I've updated the patch to remove duplicated code and actually use the new `_cmf_route` parameter

src/ChainRouterGenerateBcLayer.php
acrobat: Correct, fixed!

composer.json
```diff
@@ -29,7 +29,7 @@
         "symfony-cmf/testing": "^3@dev"
     },
     "suggest": {
-        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^3.4 || ^4.3)"
+        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^3.4 || ^4.3 || ^5.0‡ )"
```
dbu: good find! there is some weird character after the 5.0, can you remove that character?

src/ContentAwareGenerator.php
```diff
@@ -68,7 +68,9 @@ public function setContentRepository(ContentRepositoryInterface $contentReposito
      */
     public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
     {
-        if ($name instanceof SymfonyRoute) {
+        if (array_key_exists('_cmf_route', $parameters) && $parameters['_cmf_route'] instanceof SymfonyRoute) {
+            $route = $this->getBestLocaleRoute($parameters['_cmf_route'], $parameters);
+        } elseif ($name instanceof SymfonyRoute) {
```
dbu: to avoid edge case BC breaks, i would prefer to take the `$name instanceof ...` line as the first condition.

src/ChainRouterGenerateBcLayer.php
```diff
@@ -0,0 +1,127 @@
...
+        /**
+         * Loops through all registered routers and returns a router if one is found.
+         * It will always return the first route generated.
+         */
+        public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
+        {
+            if (!is_string($name)) {
+                @trigger_error('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "_cmf_route" parameter in the parameters array.', E_USER_DEPRECATED);
+
+                $parameters['_cmf_route'] = $name;
```
dbu: lets use RouteObjectInterface::ROUTE_OBJECT (but only overwrite if that field is not already set, avoid overwriting it)

src/ChainRouter.php
```diff
@@ -30,7 +29,7 @@
  * @author Henrik Bjornskov <henrik@bjrnskov.dk>
  * @author Magnus Nordlander <magnus@e-butik.se>
  */
-class ChainRouter implements ChainRouterInterface, WarmableInterface
+class ChainRouter extends ChainRouterGenerateBcLayer implements ChainRouterInterface, WarmableInterface
```
dbu: we could have the if/else for the ChainRouter, and have the ChainRouterBcLayer provide the protected methods and all, but have the actual signature difference in this class. that would eliminate one level of nested inheritance.

src/DynamicRouter.php
```diff
@@ -259,8 +230,10 @@ public function match($pathinfo)
     public function matchRequest(Request $request)
     {
         if ($this->eventDispatcher) {
+            $eventName = Events::PRE_DYNAMIC_MATCH_REQUEST;
```
dbu: lets inline the constant into the `doDispatch` call

src/DynamicRouterGenerateBcLayer.php
```diff
@@ -0,0 +1,87 @@
...
+            $routeParam = $name;
+            if (array_key_exists('_cmf_route', $parameters) && is_object($parameters['_cmf_route'])) {
+                $routeParam = $parameters['_cmf_route'];
+            }
+
+            $eventName = Events::PRE_DYNAMIC_GENERATE;
+            $event = new RouterGenerateEvent($routeParam, $parameters, $referenceType);
+
+            // LegacyEventDispatcherProxy exists in Symfony >= 4.3
+            if (class_exists(LegacyEventDispatcherProxy::class)) {
```
dbu: lets have the Bc layer class provide the protected `doDispatch` method to be used here and by the actual DynamicRouter class

src/ChainRouterGenerateBcLayer.php
```diff
@@ -0,0 +1,127 @@
...
+$refl = new \ReflectionClass(UrlGeneratorInterface::class);
+$generateMethod = $refl->getMethod('generate');
+$methodParameters = $generateMethod->getParameters();
+/** @var \ReflectionParameter $nameParameter */
+$nameParameter = array_shift($methodParameters);
+if ($nameParameter && $nameParameter->hasType() && 'string' === $nameParameter->getType()) {
+    /**
+     * @internal
+     */
+    class ChainRouterGenerateBcLayer extends ChainRouterBaseBcLayer
```
dbu: why do you call this class "Generate"? 

if we flip things around as proposed above, we could call the base class simply ChainRouterBcLayer, wdyt?

src/Event/RouterGenerateEvent.php
```diff
@@ -24,7 +23,7 @@
  *
  * @see \Symfony\Component\Routing\Generator\UrlGeneratorInterface::generate()
  */
-class RouterGenerateEvent extends Event
+class RouterGenerateEvent extends BcEvent
```
dbu: instead of a new class and file, we could use [class_alias](https://www.php.net/manual/de/function.class-alias.php) before the class declaration, wdyt?

```
if (is_subclass_of(EventDispatcherInterface::class, ContractsEventDispatcherInterface::class)) {
    \class_alias(ContractsEvent::class, 'Event');
} else {
    \class_alias(LegacyEvent::class, 'Event');
}
```

src/ChainRouter.php
acrobat: Wouldn't it be better to keep this construction so that we an just remove the bc layer class as we drop sf <5 support? Original class stays the way it is and the generate method is moved back and the bc class is removed. Or am I misunderstanding your comment?

composer.json
acrobat: Good catch! Fixed!

src/DynamicRouter.php
acrobat: Good point, fixed for both parameters and both `doDispatch` calls!

src/DynamicRouterGenerateBcLayer.php
acrobat: Good idea!

src/Event/RouterGenerateEvent.php
acrobat: I don't think this works because of the 2 event classes, I get this error if I add this to both events

```
Cannot declare class Symfony\Cmf\Component\Routing\Event\Event, because the name is already in use
```

src/ContentAwareGenerator.php
acrobat: Done! I've also switch the condition back in the `ProviderBasedGenerator`

src/ChainRouterGenerateBcLayer.php
acrobat: Indeed the naming could be better!

src/ChainRouterGenerateBcLayer.php
acrobat: All usages have been updated!

src/Event/RouterGenerateEvent.php
dbu: hm, there should be no class Event (only Events) in that namespace. but we could name the thing BaseEvent maybe?

dbu: ah crap, i think now i understood what you say. yeah, we have two events that need this. they would need to check if the other already created the alias - this is becoming a mess, lets keep your solution.

src/ContentAwareGenerator.php
dbu: i think you switched it below in getRouteByContent, but not here on this line

src/ChainRouter.php
dbu: i think both have up- and downsides (the diff is probably more readable with what i suggested), but there is less fiddling around with the final class with your approach.

you are doing the work, i let you chose ;-)

dbu:
thanks, this is shaping up nicely!

what do you think about:

> the dynamic router of the bundle further puts that from the route resolution into the parameters - unfortunately with a different key: https://github.com/symfony-cmf/routing-bundle/blob/master/src/Routing/DynamicRouter.php (i think back in the days we wanted to mimik the symfony behaviour of having underscore names in the defaults, but camelCase in the request attributes). i am a bit unsure if we should be looking for both keys, or only for the one from routing-bundle (which would require us to port the constant to the component so we can use it).

src/ContentAwareGenerator.php
acrobat: Yes indeed I missed that one, fixed now!

src/ChainRouter.php
acrobat: I'm personally more in favor of the current setup, is a bit more clear that all the bc stuff is in the separate class and it is easier to clean up afterwards

acrobat:
> thanks, this is shaping up nicely!
> 
> what do you think about:
> 
> > the dynamic router of the bundle further puts that from the route resolution into the parameters - unfortunately with a different key: https://github.com/symfony-cmf/routing-bundle/blob/master/src/Routing/DynamicRouter.php (i think back in the days we wanted to mimik the symfony behaviour of having underscore names in the defaults, but camelCase in the request attributes). i am a bit unsure if we should be looking for both keys, or only for the one from routing-bundle (which would require us to port the constant to the component so we can use it).

But if I see it correctly the dynamic router of the bundle uses the constant we are using too, so it will actually be ok the way it is? Or am I missing something?

https://github.com/symfony-cmf/routing-bundle/blob/ec8cb976f50ddf7c792b6cf18d87259b914d1c50/src/Routing/DynamicRouter.php#L91-L94

acrobat:
@dbu are the extra changes needed for your previous comment/question? Or any other remarks on the PR?

dbu:
i was AFK for the weekend, hence the slow response.

> But if I see it correctly the dynamic router of the bundle uses the constant we are using too, so it will actually be ok the way it is? Or am I missing something?

from how i read that section, the router moves the attribute from "our" key to its own key, meaning that when you generate, the attribute would be at the key the bundle specifies. we could make it stop doing that (copying rather than moving, for BC), or we could make this PR look at both keys. i feel that looking at both keys is less likely to cause trouble for existing systems, but it is the less elegant solution. maybe we should do both: no longer unset the $default and announce that the next major of the bundle will operate on the RouteObjectInterface::ROUTE_OBJECT and stop using self::ROUTE_KEY, while for now have this component look at both keys, prefering ROUTE_KEY when it is found. (as people working with the bundle might be updating that key to achieve some effect. people using only the component won't set the ROUTE_KEY but work on ROUTE_OBJECT exclusively.)

pableu:
What's needed for this to go forward? We'd love to be able to upgrade our project using symfony-cmf/Routing to Symfony 5. It this the only piece of the puzzle needed to solve #240 and symfony-cmf/routing-bundle#447 too?

dbu:
@pableu if you don't use the bundle, then this MR should be all you need. if you use the bundle, the bundle MR will be needed too, but i expect that one to be relatively simple once this thing is solved.

if you have inputs or opinions on how we should solve the parameter naming here, i would be glad to hear them. the problem is having both a reasonable thing for the future but not break BC if possible.

webhdx:
It's a pity there is hardly any traction in this PR. There are lots of projects upgrading to Symfony 5 and this package is a huge blocker. At least for us at ezsystems/ezplatform.

@dbu I think this is a good approach but I'd prefer to have a proper route name indicating it will use an object from parameters instead of an empty string.

Is this the last issue before you could tag next major with Symfony 5 support? I could help if  @acrobat is no longer interested in finishing it.

acrobat:
I think we’re almost finished here, except for @dbu’s last comment. I will try to find some time to finish this pr.

dbu:
> I think this is a good approach but I'd prefer to have a proper route name indicating it will use an object from parameters instead of an empty string.

that is an interesting idea. that would also allow us to have a BC behaviour and the new behaviour only if that name is used. the name should be fairly specific to avoid accidental clashes with other routers or even actual route names. something like `cmf_routing_object` maybe?

alexander-schranz:
@webhdx idea sounds great 👍 . @dbu `cmf_routing_object` sounds good for me.

webhdx:
`cmf_routing_object` is fine for me as well :+1:

acrobat:
So how to continue with this, replace the current usages of `RouteObjectInterface::ROUTE_OBJECT` to a new key?

src/ProviderBasedGenerator.php
```diff
@@ -49,6 +49,8 @@ public function generate($name, $parameters = [], $referenceType = self::ABSOLUT
     {
         if ($name instanceof SymfonyRoute) {
             $route = $name;
+        } elseif ('' === $name && array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && $parameters[RouteObjectInterface::ROUTE_OBJECT] instanceof SymfonyRoute) {
```
dbu: i think that using RouteObjectInterface::ROUTE_OBJECT as key is fine here. we would use the new magic route name to trigger this, instead of `''`. please define as a constant in the RouteObjectInterface.

acrobat: So using `cmf_routing_object` instead of an empty string?

webhdx: Yes this is an idea. Just create new constant in `RouteObjectInterface` i.e. `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME` and use it instead of an empty string.

acrobat: aight 👍

acrobat:
I've pushed an update with the specific routename

src/ChainRouterBcLayer.php
```diff
@@ -0,0 +1,127 @@
...
+    class ChainRouterBcLayer extends ChainRouterBaseBcLayer
+    {
+        /**
+         * Loops through all registered routers and returns a router if one is found.
+         * It will always return the first route generated.
+         */
+        public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
+        {
+            if (!is_string($name)) {
+                @trigger_error(sprintf('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "%s" parameter in the parameters array.', RouteObjectInterface::ROUTE_OBJECT), E_USER_DEPRECATED);
```
webhdx: Error message needs to be changed as well @acrobat

src/DynamicRouterBcLayer.php
```diff
@@ -0,0 +1,92 @@
...
+} else {
+    /**
+     * @internal
+     */
+    class DynamicRouterBcLayer extends DynamicRouterBaseBcLayer
+    {
+        public function generate($name, $parameters = [], $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
+        {
+            if (!is_string($name)) {
+                @trigger_error(sprintf('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass an empty route name and the object as "%s" parameter in the parameters array.', RouteObjectInterface::ROUTE_OBJECT), E_USER_DEPRECATED);
```
webhdx: Same

src/ChainRouterBcLayer.php
acrobat: Good catch, updated!

src/DynamicRouterBcLayer.php
acrobat: Good catch, updated!

acrobat:
@dbu I think this is ready for a final review!

src/ChainRouterBcLayer.php
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+namespace Symfony\Cmf\Component\Routing;
+
+use Symfony\Component\Routing\Exception\RouteNotFoundException;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+$refl = new \ReflectionClass(UrlGeneratorInterface::class);
+$generateMethod = $refl->getMethod('generate');
+$methodParameters = $generateMethod->getParameters();
```
dbu: there are all global variables, which has me worried a bit. for that reason, i would prefer to rewrite this as

```
$urlGeneratorNameParameter = array_shift((new \ReflectionClass(UrlGeneratorInterface::class)->getMethod('generate')->getParameters());
```

src/DynamicRouterBcLayer.php
```diff
@@ -0,0 +1,92 @@
...
+use Symfony\Cmf\Component\Routing\Event\Events;
+use Symfony\Cmf\Component\Routing\Event\RouterGenerateEvent;
+use Symfony\Component\EventDispatcher\LegacyEventDispatcherProxy;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+
+$refl = new \ReflectionClass(UrlGeneratorInterface::class);
+$generateMethod = $refl->getMethod('generate');
+$methodParameters = $generateMethod->getParameters();
+/** @var \ReflectionParameter $nameParameter */
+$nameParameter = array_shift($methodParameters);
```
dbu: same here, lets have only one global variable and lets have that one have a more specific name

src/RouteObjectInterface.php
```diff
@@ -56,6 +56,11 @@ interface RouteObjectInterface
      */
     const CONTENT_ID = '_content_id';
 
+    /**
+     * Routename used when passing a route object to the generator.
```
dbu:
```suggestion
     * Route name used when passing a route object to the generator in $parameters[RouteObjectInterface::ROUTE_OBJECT].
```

src/DynamicRouterBcLayer.php
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @internal
+ */
+abstract class DynamicRouterBaseBcLayer
+{
+    protected function doGenerate($name, $parameters, $referenceType)
+    {
+        if ($this->eventDispatcher) {
+            $routeParam = $name;
+            if (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
```
dbu: to be strict BC, we should first check if `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name` and only then use the object from the parameters

src/DynamicRouterBcLayer.php
```diff
@@ -0,0 +1,92 @@
...
+     */
+    class DynamicRouterBcLayer extends DynamicRouterBaseBcLayer
+    {
+        public function generate($name, $parameters = [], $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
+        {
+            if (!is_string($name)) {
+                @trigger_error(sprintf('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass the `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME` constant as the route name and the object as "%s" parameter in the parameters array.', RouteObjectInterface::ROUTE_OBJECT), E_USER_DEPRECATED);
+
+                if (!isset($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
+                    $parameters['_cmf_route'] = $name;
```
dbu: I think this should be:

```suggestion
                    $parameters[RouteObjectInterface::ROUTE_OBJECT] = $name;
```

but doing this is a BC break for people having their own things hooked with the dynamic router. i would only trigger the deprecation warning but not rewrite the parameters. the `doGenerate` still handles the old way with the object as $name, so we don't need to change this here.

src/ContentAwareGenerator.php
```diff
@@ -70,6 +70,8 @@ public function generate($name, $parameters = [], $absolute = UrlGeneratorInterf
     {
         if ($name instanceof SymfonyRoute) {
             $route = $this->getBestLocaleRoute($name, $parameters);
+        } elseif (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && $parameters[RouteObjectInterface::ROUTE_OBJECT] instanceof SymfonyRoute) {
```
dbu: i think for strict BC, we need here (and in getRouteByContent) to check `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name`. otherwise if someone happened to have the route object in the parameters, they see an unexpected change.

dbu: i think for strict BC, we need here (and in getRouteByContent) to check `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name`. otherwise if someone happened to have the route object in the parameters, they see an unexpected change.

src/ChainRouterBcLayer.php
acrobat: I will need 1 more variable here as php will throw a notice "Only variables should be passed by reference". This will also make the tests fail

src/DynamicRouterBcLayer.php
acrobat: I actually need to keep doing this as the doGenerate call with $name being an object will fail on symfony 5 (typehint)

src/ContentAwareGenerator.php
acrobat: Fixed

acrobat: Fixed

src/DynamicRouterBcLayer.php
acrobat: Fixed

src/RouteObjectInterface.php
acrobat: Fixed

src/DynamicRouterBcLayer.php
acrobat: See my comment above, but this is fixed I think!

acrobat:
@dbu I've addressed your comments except for the comment about the doGenerate calm

src/DynamicRouterBcLayer.php
```diff
@@ -0,0 +1,87 @@
...
+     */
+    class DynamicRouterBcLayer extends DynamicRouterBaseBcLayer
+    {
+        public function generate($name, $parameters = [], $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH)
+        {
+            if (!is_string($name)) {
+                @trigger_error(sprintf('Passing an object as the route name is deprecated in symfony-cmf/Routing v2.2 and will not work in Symfony 5.0. Pass the `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME` constant as the route name and the object as "%s" parameter in the parameters array.', RouteObjectInterface::ROUTE_OBJECT), E_USER_DEPRECATED);
+
+                $parameters[RouteObjectInterface::ROUTE_OBJECT] = $name;
+                $name = RouteObjectInterface::OBJECT_BASED_ROUTE_NAME;
```
dbu: i still don't think we need to rewrite here.

when we are on symfony 5, we have the other variant of the DynamicRouterBcLayer defined above, that simply forwards the generate call. if we are pre symfony 5, we can call doGenerate with $name being an object. doGenerate does not declare the argument type, and calls Generator::generate() which will not have the string declaration if its symfony 4.

src/DynamicRouterBcLayer.php
```diff
@@ -0,0 +1,87 @@
...
+        if ($this->eventDispatcher) {
+            $routeParam = $name;
+            if (RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name && array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
+                $routeParam = $parameters[RouteObjectInterface::ROUTE_OBJECT];
+            }
+
+            $event = new RouterGenerateEvent($routeParam, $parameters, $referenceType);
+            $this->doDispatch(Events::PRE_DYNAMIC_GENERATE, $event);
+
+            $name = $event->getRoute();
```
dbu: hm, together with the code above, this will end up putting the route object into the name, no? i think we are missing a test that would reveil the problem.

i think the intention is that the event can alter what to generate. we could forward the parameters into the event class as we did before and not read the route object from the parameters. it will be the job of the generator to do that.

additionally, we should adjust the event class https://github.com/symfony-cmf/Routing/blob/098ea1d035e2d2bda804e330fb99d39d1726123e/src/Event/RouterGenerateEvent.php#L64 setter, getter and constructor to say that the route should be a string and having objects will start failing in symfony 5.

src/DynamicRouterBcLayer.php
acrobat: Hmm you are right I think, let look into it!

src/DynamicRouterBcLayer.php
acrobat: @dbu I've added deprecations in the event class and a test to verify it.

But I'm not sure if I understand what you mean with

> i think the intention is that the event can alter what to generate. we could forward the parameters into the event class as we did before and not read the route object from the parameters. it will be the job of the generator to do that.

Can you explain it a bit more? I think this is the last piece to finish this PR!

src/DynamicRouterBcLayer.php
acrobat: You were correct, I've removed these lines!

alexander-schranz:
I'm looking forward to this. @dbu is anything missing here?

src/DynamicRouterBcLayer.php
dbu: sorry for the late reply, instead of lots of time for OS, corona lockdown made me have not much time at all...

the idea of the event is to have a place where users can change parameters or even the route that will be generated. i think we should create that event with $name and $parameters like before, and remove the `$routeParam = $parameters[RouteObjectInterface::ROUTE_OBJECT];` part. if we get there, we create the event with a route object and then the line this comment is attached to will put the object into `$name` which we then pass to the generate method, which violates the restriction that $name must be string.

imho we should dispatch the event with unmodified `$name` rather than put the route object into $name.

gisostallenberg:
> I'm looking forward to this. @dbu is anything missing here?

Same here, @acrobat and @dbu can I do something to help?

src/ChainRouterBcLayer.php
```diff
@@ -0,0 +1,124 @@
...
+                continue;
+            }
+
+            $routeName = $name;
+            if (RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name && array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
+                $routeName = $parameters[RouteObjectInterface::ROUTE_OBJECT];
+            }
+
+            // If $router is versatile and doesn't support this route name, continue
+            if ($router instanceof VersatileGeneratorInterface && !$router->supports($routeName)) {
```
dbu: i think here we should pass in the route object if existing. we would need to change the phpdoc of VersatileRouterInterface, but that interface explicitly says that the argument could be an object. then routers can decide on the type of the route_object parameter whether they support generating a route for that.

dbu:
> Same here, @acrobat and @dbu can I do something to help?

each time i look over the diff, i find something new that would lead to a BC break. its really quite tricky here, to not break existing applications.

things that would help
* review the code and try to think if existing code might break from the behaviour change if it is not changed, or if it is changed to only use route names and the parameters for the route object. that includes the event listeners and everything. i don't think it will be possible to solve everything here, particularly custom route generators that rely on the object being in $name.
* try this branch in your application for a sanity check

one option would be to instead of this make a hard break with a new major version to keep things a bit simpler - without BC breaks, there are just so many hackeries we need to do to not break any existing code.

dbu:
> each time i look over the diff, i find something new that would lead to a BC break. 

@acrobat that was not a slant at you, btw. i appreciate the effort you put into this, and the patience with my feedback. its just that i am afraid that we attempt something impossible with keeping BC with such a fundamental change to the basic assumption of this codebase.

gisostallenberg:
> one option would be to instead of this make a hard break with a new major version to keep things a bit simpler - without BC breaks, there are just so many hackeries we need to do to not break any existing code.

What would the upgrade path look like if you did this? Would it just be 'check your code for objects passed as names, choose a name and pass the object in the parameters'? In that case I'd suggest avoid adding all this complex BC layers which also needs to be maintained. But that is just an opinion :wink:. I think you and @acrobat are doing an incredible job in trying to move forward and it is much appreciated, but maybe it is just to hard to keep BC and it would be easy to upgrade a library using your package...

dbu:
yep, basically the upgrade path will be to move route objects into the parameters, if something is calling route generation explicitly. the tricky bits will be listeners and event handlers and other places where this change is rather implicit.

if we don't do this whole forward/backward compatible thing, we can't really add the deprecations though - deprecating $name be an object when there is no other way to generate the route would be weird.

if we would aim to support the new and the old way but without any magic to work with symfony 5, we would probably still need at least 50% of the changes in here. which is still quite some effort, with the risk of breaking things...

i tend to say lets just make a BC breaking new major version that only works with symfony 5 and a rather hard break for it.

gisostallenberg:
Although it is a pity all hard work seems te have been for nothing then, I agree with @dbu.
What do you think @acrobat ?

acrobat:
For me it's either solution. If we want to keep trying I'm up for it or if we choose to do the hard BC-break also good for me!

Maybe we should still try to add some deprecation notices about the change for object route names?

dbu:
> Maybe we should still try to add some deprecation notices about the change for object route names?

if we don't support the new way, we should not do deprecations messages, as there would be no way to fix them.

acrobat:
Good point! So what would be the next step here? Create a new PR and strip out all the routename as object support code and create a pr for 2.x with a small upgrade guide?

What will this mean for the RoutingBundle? Also a new major version there?

dbu:
yes. sorry acrobat for not realizing earlier :-( your effort is not wasted though, we can use it to build the symfony 5 compatible version, and the rest was the proof that we can't reasonably support 4 and 5 at the same time...

master is 2.2. i suggest you start a new branch against master where you bump branch-alias to 3.x and to the BC breaks as needed to only have the new way. i can then create the 2.2 branch right before merging.

acrobat:
No worries @dbu! It was a search for all of us.

I will start on a new pr!

acrobat:
I've made a start for the new approach in #248

dbu:
finished in #250

PR #248, opened 2020-04-02
Symfony 5 support

acrobat:
| Q             | A
| ------------- | ---
| Branch?       | "master"
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

As discussed in #245 this PR adds symfony support without the required bc layer for objects as routeName. I've bumped the php version to 7.2 to allow us to keep the current generator signature for now, which allows for an easier upgrade. 

@dbu should I open a PR for the dev-kit to bump the php version tested in travis?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.2-dev"
+            "dev-master": "2.x-dev"
         }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -15,20 +15,20 @@
         }
     ],
     "require": {
-        "php": "^7.1",
-        "symfony/routing": "^3.4 || ^4.3",
-        "symfony/http-kernel": "^3.4 || ^4.3",
+        "php": "^7.2",
+        "symfony/routing": "^4.4 || ^5.0",
+        "symfony/http-kernel": "^4.4 || ^5.0",
         "psr/log": "^1.0"
     },
     "require-dev": {
-        "symfony/phpunit-bridge": "^4.2.2",
-        "symfony/dependency-injection": "^3.4 || ^4.3",
-        "symfony/config": "^3.4 || ^4.3",
-        "symfony/event-dispatcher": "^3.4 || ^4.3",
+        "symfony/phpunit-bridge": "^5.0",
+        "symfony/dependency-injection": "^4.4 || ^5.0",
+        "symfony/config": "^4.4 || ^5.0",
+        "symfony/event-dispatcher": "^4.4 || ^5.0",
         "symfony-cmf/testing": "^3@dev"
     },
     "suggest": {
-        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^3.4 || ^4.3)"
+        "symfony/event-dispatcher": "DynamicRouter can optionally trigger an event at the start of matching. Minimal version (^4.4 || ^5.0)"
     },
     "autoload": {
         "psr-4": {
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -218,6 +218,10 @@ class ChainRouter implements ChainRouterInterface, WarmableInterface
      */
     public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
     {
+        if (is_object($name)) {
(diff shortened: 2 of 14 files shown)


composer.json
```diff
@@ -15,20 +15,20 @@
         }
     ],
     "require": {
-        "php": "^7.1",
-        "symfony/routing": "^3.4 || ^4.3",
-        "symfony/http-kernel": "^3.4 || ^4.3",
+        "php": "^7.2",
+        "symfony/routing": "^3.4 || ^4.3 || ^5.0",
```
dbu, line 19: i suggest that we restrict oursevles to `^4.4 || ^5.0`. 4.3 is fading out, and 3.4 users can use version 2 of this library.

dbu:
> should I open a PR for the dev-kit to bump the php version tested in travis?

once we are ready to merge, we should adjust the dev-kit. before that, editing the file by hand is the way to go.

acrobat:
@dbu php 7.2 has type widening implemented (https://wiki.php.net/rfc/parameter-no-type-variance). This allows to drop the typehint on implenting classes, so indeed this could help us in v2.x but would it solve the whole BC breaking issue we try to avoid?

acrobat:
> > should I open a PR for the dev-kit to bump the php version tested in travis?
> 
> once we are ready to merge, we should adjust the dev-kit. before that, editing the file by hand is the way to go.

Ok, will do!

composer.json
acrobat: Good 👍

dbu:
> @dbu php 7.2 has type widening implemented (https://wiki.php.net/rfc/parameter-no-type-variance). This allows to drop the typehint on implenting classes, so indeed this could help us in v2.x but would it solve the whole BC breaking issue we try to avoid?

i think it would. version 2.3 would simply implement looking in the parameters but falling back to $name being the object if the parameter is not there, but raise a deprecation warning if the $name is not a string. i would let each bit check that separately and not have e.g. the chain router convert from old to new - the chain router, the event and the generator would each need to support the new mode while falling back to the old mode.

acrobat:
That would indeed be possible! Let me see where I can come up with, I will tag you if you can review the new solution!

acrobat:
I've pushed an update where I've added the deprecation notice and the replacement to pass objects into the generate method. 

Next up is added some tests for the new object parameter, I think this will also highlight some area's we still need to update

alexander-schranz:
Looks like travis did just don't get any feedback but the tests are green. StyleCI seems to have some issues which need to be fixed.

src/ChainRouter.php
```diff
@@ -227,8 +231,19 @@ public function generate($name, $parameters = [], $absolute = UrlGeneratorInterf
                 continue;
             }
 
+            // if $router does not announce it is capable of handling
+            // non-string routes and the ROUTE_OBJECT is set in the parameters array, continue
+            if (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT]) && !$router instanceof VersatileGeneratorInterface) {
```
dbu, line 236: i would not do this check. we can't know how the router wants to work, and if the route name is our "magic" route, the router will simply not find a route to generate. (see my other comment below)

src/ChainRouter.php
```diff
@@ -227,8 +231,19 @@ public function generate($name, $parameters = [], $absolute = UrlGeneratorInterf
...
+                continue;
+            }
+
+            $routeName = $name;
+            if (RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name && array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters) && is_object($parameters[RouteObjectInterface::ROUTE_OBJECT])) {
+                $routeName = $parameters[RouteObjectInterface::ROUTE_OBJECT];
```
dbu, line 242: i would not do this. instead i would only call the versatile generator `supports` if $name is an object, and deprecate the interface. we had this because normal routers choke on $name being an object. with the new way of doing this, $name won't be an object anymore and the router can just do a 404.

src/ChainRouter.php
acrobat: Good point, I will revert this!

acrobat:
@dbu I think I fixed most of the comments, so I guess this PR is ready for another review round!

dbu:
looks good, i am doing a couple of tweaks now...

PR #251, opened 2020-05-04
have provider based generator support new mode too

dbu:
| Q             | A
| ------------- | ---
| Branch?       | "master"
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/ContentAwareGenerator.php
+++ b/src/ContentAwareGenerator.php
@@ -80,7 +80,9 @@ class ContentAwareGenerator extends ProviderBasedGenerator
 
             $route = $this->getBestLocaleRoute($name, $parameters);
         } elseif (RouteObjectInterface::OBJECT_BASED_ROUTE_NAME === $name) {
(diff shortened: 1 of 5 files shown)


src/ContentAwareGenerator.php
```diff
@@ -98,8 +100,9 @@ public function generate($name, $parameters = [], $absolute = UrlGeneratorInterf
         }
 
         $this->unsetLocaleIfNotNeeded($route, $parameters);
+        $parameters[RouteObjectInterface::ROUTE_OBJECT] = $route;
 
-        return parent::generate($route, $parameters, $absolute);
+        return parent::generate(RouteObjectInterface::OBJECT_BASED_ROUTE_NAME, $parameters, $absolute);
```
dbu, line 105: we need to switch this around. this is inheritance, not composition, so there should not be any unexpected side effects as with the chain router.

dbu:
@acrobat i realized that we forgot to adjust the ProviderBasedGenerator. could you please review this PR that adds it there, and adjusts the ContentAwareGenerator to not always trigger the deprecation?

PR #253, opened 2020-05-27
fix passing route object parameter to generator

gseidel:
| Q             | A
| ------------- | ---
| Branch?       | 2.1
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

If the route is fetched over a relation from the database, it could be, that the route is still a proxy object (For lazy loading). If a url is generated with a proxy route object, then symfony will generate a string like ``/path?_route_object%5B__isInitialized__%5D=1``, because ``http_build_query`` also accept objects and generate a value if the object has a public property.

I think the parameter ``RouteObjectInterface::ROUTE_OBJECT`` is not necessary to pass it ``doGenerate``. So it should be unset if exists.


gseidel:
Wrong branch. Need to reopen on master

PR #254, opened 2020-05-27
fix passing route object parameter to generator

gseidel:
| Q             | A
| ------------- | ---
| Branch?       | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

If the route is fetched over a relation from the database, it could be, that the route is still a proxy object (For lazy loading). If a url is generated with a proxy route object, then symfony will generate a string like ``/path?_route_object%5B__isInitialized__%5D=1``, because ``http_build_query`` also accept objects and generate a value if the object has a public property.

I think the parameter ``RouteObjectInterface::ROUTE_OBJECT`` is not necessary to pass it ``doGenerate``. So it should be unset if exists.

### DIFF
--- a/src/ProviderBasedGenerator.php
+++ b/src/ProviderBasedGenerator.php
@@ -69,6 +69,11 @@ class ProviderBasedGenerator extends UrlGenerator implements VersatileGeneratorI
             throw new RouteNotFoundException(sprintf('Route "%s" does not exist.', $name));
         }
 
+        // the parameter must be unset to avoid unexpected generation behaviour
+        if (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters)) {
+            unset($parameters[RouteObjectInterface::ROUTE_OBJECT]);
+        }
+
         // the Route has a cache of its own and is not recompiled as long as it does not get modified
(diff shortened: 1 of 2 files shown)


src/ProviderBasedGenerator.php
```diff
@@ -69,6 +69,11 @@ public function generate($name, $parameters = [], $referenceType = self::ABSOLUT
             throw new RouteNotFoundException(sprintf('Route "%s" does not exist.', $name));
         }
 
+        // the parameter must be unset to avoid unexpected generation behaviour
+        if (array_key_exists(RouteObjectInterface::ROUTE_OBJECT, $parameters)) {
+            unset($parameters[RouteObjectInterface::ROUTE_OBJECT]);
```
dbu, line 74: to avoid any possible side effects, i would move this into the `elseif` right after line 67.

gseidel: Good point. I updated the code and use it inside the ``elseif``

dbu:
i updated the changelog and tagged 2.3.2 with this.

gseidel:
Thanks a lot @dbu

PR #259, opened 2020-10-03
Remove calls to deprecated TestCase::at()

andypost:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | comma-separated list of tickets fixed by the PR, if any
| License       | MIT
| Doc PR        | https://github.com/sebastianbergmann/phpunit/issues/4297

### DIFF
--- a/tests/Unit/DependencyInjection/Compiler/RegisterRouteEnhancersPassTest.php
+++ b/tests/Unit/DependencyInjection/Compiler/RegisterRouteEnhancersPassTest.php
@@ -30,7 +30,7 @@ class RegisterRouteEnhancersPassTest extends TestCase
 
         $builder = $this->createMock(ContainerBuilder::class);
         $definition = new Definition('router');
-        $builder->expects($this->at(0))
+        $builder->expects($this->atLeastOnce())
             ->method('hasDefinition')
             ->with('cmf_routing.dynamic_router')
             ->will($this->returnValue(true))
--- a/tests/Unit/Routing/PagedRouteCollectionTest.php
+++ b/tests/Unit/Routing/PagedRouteCollectionTest.php
@@ -47,12 +47,16 @@ class PagedRouteCollectionTest extends TestCase
         }
         $names = array_keys($routes);
 
+        $with = [];
+        $will = [];
         foreach ($expectedCalls as $i => $range) {
-            $this->routeProvider->expects($this->at($i))
-              ->method('getRoutesPaged')
-              ->with($range[0], $range[1])
-              ->will($this->returnValue(array_slice($routes, $range[0], $range[1])));
+            $with[] = [$range[0], $range[1]];
+            $will[] = array_slice($routes, $range[0], $range[1]);
         }
+        $mocker = $this->routeProvider->expects($this->exactly(\count($expectedCalls)))
+            ->method('getRoutesPaged');
+        \call_user_func_array([$mocker, 'withConsecutive'], $with);
+        \call_user_func_array([$mocker, 'willReturnOnConsecutiveCalls'], $will);
 
         $route_collection = new PagedRouteCollection($this->routeProvider, $routesLoadedInParallel);
 
--- a/tests/Unit/Routing/ChainRouterTest.php
+++ b/tests/Unit/Routing/ChainRouterTest.php
@@ -130,22 +130,12 @@ class ChainRouterTest extends TestCase
             ->setMethods(['sortRouters'])
             ->getMock();
         $router
-            ->expects($this->at(0))
+            ->expects($this->exactly(2))
             ->method('sortRouters')
-            ->will(
-                $this->returnValue(
-                    [$high, $medium, $low]
-                )
-            )
-        ;
-        // The second time sortRouters() is called, we're supposed to get the newly added router ($highest)
-        $router
-            ->expects($this->at(1))
-            ->method('sortRouters')
-            ->will(
-                $this->returnValue(
-                    [$highest, $high, $medium, $low]
-                )
+            ->willReturnOnConsecutiveCalls(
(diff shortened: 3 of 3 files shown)


tests/Unit/Routing/PagedRouteCollectionTest.php
```diff
@@ -47,12 +47,18 @@ public function testIterator($amountRoutes, $routesLoadedInParallel, $expectedCa
...
-              ->method('getRoutesPaged')
-              ->with($range[0], $range[1])
-              ->will($this->returnValue(array_slice($routes, $range[0], $range[1])));
+            $with[] = [$range[0], $range[1]];
+            $will[] = array_slice($routes, $range[0], $range[1]);
+            ++$counter;
         }
+        $mocker = $this->routeProvider->expects($this->exactly($counter))
+            ->method('getRoutesPaged');
+        \call_user_func_array([$mocker, 'withConsecutive'], $with);
```
dbu, line 60: if i am not mistaken, we can use `$this->routeProvider` here. `$mocker` is just a reference to the same object, as the phpunit methods return the object itself to allow chaining. can you please remove the extra variable? makes the code easier to read imho.

tests/Unit/Routing/PagedRouteCollectionTest.php
```diff
@@ -47,12 +47,18 @@ public function testIterator($amountRoutes, $routesLoadedInParallel, $expectedCa
...
         foreach ($expectedCalls as $i => $range) {
-            $this->routeProvider->expects($this->at($i))
-              ->method('getRoutesPaged')
-              ->with($range[0], $range[1])
-              ->will($this->returnValue(array_slice($routes, $range[0], $range[1])));
+            $with[] = [$range[0], $range[1]];
+            $will[] = array_slice($routes, $range[0], $range[1]);
+            ++$counter;
         }
+        $mocker = $this->routeProvider->expects($this->exactly($counter))
```
dbu, line 58: we could also `\count($expectedCalls)` instead of having a separate counter. i would find that a bit more straightforward to read, if you don't mind changing it.

tests/Unit/DependencyInjection/Compiler/RegisterRouteEnhancersPassTest.php
```diff
@@ -30,7 +30,7 @@ public function testRouteEnhancerPass()
 
         $builder = $this->createMock(ContainerBuilder::class);
         $definition = new Definition('router');
-        $builder->expects($this->at(0))
+        $builder->expects($this->atLeastOnce())
```
dbu, line 33: weird that we had this `at(0)` here. do the tests succeed with `$this->once()`? that would seem more accurate

dbu (commented):
thanks for this! i have a few nitpicks

tests/Unit/Routing/PagedRouteCollectionTest.php
andypost: both fixed, thanks for review

tests/Unit/DependencyInjection/Compiler/RegisterRouteEnhancersPassTest.php
andypost: I still think it tests `\Symfony\Cmf\Component\Routing\DependencyInjection\Compiler\RegisterRouteEnhancersPass::process()` which must call `hasDefinition()` at least once

Also after that call few others could be executed which may call it, so the method could be executed more then once

andypost: And yes tests pass with `once()`

dbu: the old code would have returned null on futher calls. but its not that important as the implementation may call it as often as it wants.

tests/Unit/Routing/PagedRouteCollectionTest.php
```diff
@@ -47,12 +47,16 @@ public function testIterator($amountRoutes, $routesLoadedInParallel, $expectedCa
...
+        $will = [];
         foreach ($expectedCalls as $i => $range) {
-            $this->routeProvider->expects($this->at($i))
-              ->method('getRoutesPaged')
-              ->with($range[0], $range[1])
-              ->will($this->returnValue(array_slice($routes, $range[0], $range[1])));
+            $with[] = [$range[0], $range[1]];
+            $will[] = array_slice($routes, $range[0], $range[1]);
         }
+        $this->routeProvider->expects($this->exactly(\count($expectedCalls)))
```
dbu, line 56: sorry, i was wrong and you were right :man_facepalming: - can you please restore the temporary variable again, as $this->routeProvider is the proxy, and not the method mocking thingy?

tests/Unit/Routing/PagedRouteCollectionTest.php
andypost: Checked and reverted because `\PHPUnit\Framework\MockObject\Builder\InvocationMocker::method()` returns is different mocker

tests/Unit/DependencyInjection/Compiler/RegisterRouteEnhancersPassTest.php
andypost: OTOH making it ``once()` is more strict and could be better, but I've no idea which is proffered

dbu: i think that actually this is better. otherwise we test implementation details rather than the behaviour.

dbu (approved):
thanks a lot!

Issue #247, opened 2020-03-04
Redirect routes with trailing slash


EmmanuelVella:
## Subject

Hi here !

I just upgraded my app from symfony 3.4 to 4.4, and I have an issue with trailing slashes. In symfony 3.4, I used [this method](https://symfony.com/doc/3.4/routing/redirect_trailing_slash.html) which worked well.

But since symfony 4.1 and the new trailing slashes handling, I get an infinite loop when trying to match a cmf route with a trailing slash. Removing the custom redirect action works, but cmf urls with trailing slashes are not redirected anymore.

## Steps to reproduce

Use symfony 4.1+, and try to match a cmf route ending with an url with a trailing slash.

## Expected results

I should be redirected to the same url without the trailing slash.

## Actual results

I get a 404.

dbu:
hi emmanuel, long time no see :wave:

yeah, unless you store the route with the flag to have a trailing `/`, it will not be matched. if i understand https://symfony.com/doc/4.1/routing.html#routing-trailing-slash-redirection correctly, they now automatically redirect, but only if the route without a slash has been found. we could implement the same in the cmf route matcher (we already have the logic to see if [the flag](https://github.com/symfony-cmf/routing-bundle/blob/513c269adb2d72bd5daf309d343374bf396be8c2/src/Doctrine/Phpcr/Route.php#L65) that a route is supposed to have a trailing slash) - i assume the only difference is that we would need to also provide the inverse logic to redirect removing the slash if no trailing slash is wanted. the [candidates class](https://github.com/symfony-cmf/Routing/blob/12f3f64a0ab89ee0dc5ead8802167de762eac024/src/Candidates/Candidates.php#L141) has the logic to strip trailing slashes. 

i would assume that we need something in the [UrlMatcher](https://github.com/symfony-cmf/Routing/blob/master/src/NestedMatcher/UrlMatcher.php) to find out that we need to redirect, or in DynamicRouter - i am not sure at which level core symfony does the redirection, but think that we should do it on the same level.

do you want to do a pull request for that?

dannyvw:
Is there any update for this?

dbu:
i don't think anyone worked on it, no. want to give it a shot?

EmmanuelVella:
Hi all ! Sorry, I just wanted to report the issue, I would be glad to help but unfortunately I have no time for this.

dbu:
[...]

PR #265, opened 2021-06-11
handle trailing slash

johndodev:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/symfony-cmf/routing-bundle/issues/464 
|                      | https://github.com/symfony-cmf/Routing/issues/247
| License       | MIT

### DIFF
--- a/src/ChainRouter.php
+++ b/src/ChainRouter.php
@@ -18,6 +18,7 @@ use Symfony\Component\Routing\Exception\MethodNotAllowedException;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 use Symfony\Component\Routing\Exception\RouteNotFoundException;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Routing\Matcher\RedirectableUrlMatcherInterface;
 use Symfony\Component\Routing\Matcher\RequestMatcherInterface;
 use Symfony\Component\Routing\RequestContext;
 use Symfony\Component\Routing\RequestContextAwareInterface;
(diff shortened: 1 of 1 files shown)


src/ChainRouter.php
```diff
@@ -370,4 +390,17 @@ public function hasRouters()
     {
         return 0 < count($this->routers);
     }
+
+    public function redirect(string $path, string $route, string $scheme = null): array
+    {
+        return [
+            '_controller' => 'Symfony\\Bundle\\FrameworkBundle\\Controller\\RedirectController::urlRedirectAction',
```
dbu, line 397: we do not have a dependency on the framework-bundle in this component, so this won't always work.

dbu (changes requested):
thanks a lot for looking into this :+1: 

i looked a bit into the symfony code, and saw RedirectableUrlMatcher in symfony/routing as well as RedirectableCompiledUrlMatcher in the framework-bundle.

inspired by those, i suggest that we write a RedirectableUrlMatcher in the RoutingBundle repository, as there we depend on framework-bundle and can provide the redirect instructions.

that matcher would take a router as constructor argument and do the slash logic.

the bundle service configuration would then wrap DynamicRouter with the RedirectableUrlMatcher. and we can document how to use that router when registering other routers for the ChainRouter. if we wrap the whole chain router, we would retry actual "not founds" three times with the symfony router (once within with the symfony router logic, then with the added/removed slash from chain router which makes the symfony router again do 2 match attempts).

dbu:
thanks for https://github.com/symfony-cmf/routing-bundle/pull/465 !

PR #269, opened 2021-12-16
Prepare major release and allow symfony 6

acrobat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        |

This PR adds some required typehints for symfony 6 and to make this change easier I suggest to release this PR as a new major release. It's easy for project to upgrade/support the upcoming major, as only a previous deprecation is removed and all older symfony versions are still supported (+ sf6 support)

### DIFF
--- a/src/ChainRouteCollection.php
+++ b/src/ChainRouteCollection.php
@@ -43,17 +43,15 @@ class ChainRouteCollection extends RouteCollection
      *
      * @return \ArrayIterator An \ArrayIterator object for iterating over routes
      */
-    public function getIterator()
+    public function getIterator(): \ArrayIterator
     {
         return new \ArrayIterator($this->all());
     }
 
     /**
      * Gets the number of Routes in this collection.
-     *
-     * @return int The number of routes
      */
-    public function count()
+    public function count(): int
     {
         $count = 0;
         foreach ($this->routeCollections as $routeCollection) {
@@ -81,7 +79,7 @@ class ChainRouteCollection extends RouteCollection
      *
      * @return Route[] An array of routes
      */
-    public function all()
+    public function all(): array
     {
         $routeCollectionAll = new RouteCollection();
         foreach ($this->routeCollections as $routeCollection) {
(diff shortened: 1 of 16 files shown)


acrobat:
All tests pass again, the PR is ready for review!

src/ChainRouteCollection.php
```diff
@@ -98,14 +98,16 @@ public function all()
      *
      * @return Route|null A Route instance or null when not found
```
dbu, line 99:
```suggestion
```

src/ChainRouter.php
```diff
@@ -213,23 +210,13 @@ private function doMatch($pathinfo, Request $request = null)
...
+     * @param string $name
      *
      * Loops through all registered routers and returns a router if one is found.
      * It will always return the first route generated.
      */
-    public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH)
+    public function generate($name, $parameters = [], $absolute = UrlGeneratorInterface::ABSOLUTE_PATH): string
     {
-        if (is_object($name)) {
-            @trigger_error('Passing an object as route name is deprecated since version 2.3. Pass the `RouteObjectInterface::OBJECT_BASED_ROUTE_NAME` as route name and the object in the parameters with key `RouteObjectInterface::ROUTE_OBJECT`.', E_USER_DEPRECATED);
```
dbu, line 230: should we check $name with is_string and error if it is not a string?

dbu (commented):
thanks for that work!

you are removing the deprecation about passing objects - i think if we don't have `string` parameter type declarations, we should check with is_string to have clean error reporting rather than end up in unexpected behaviour.

acrobat:
I've add the necessary type checks and tests for this case

src/ChainRouter.php
acrobat: Good point, I've applied the check!

dbu (approved):
thanks for the work!

dbu:
i adjusted the branch alias in composer.json.

can you add a build for php 8.1 as well? that might lead to further changes because of deprecations.

Issue #278, opened 2022-06-30
TypeError for LazyRouteCollection::all() when inner route provider returns generator from getRoutesByNames()


andrew-demb:
#### Symfony packages

```
$ composer show --latest 'symfony/*'
symfony/asset                      v6.1.0  v6.1.0  Manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and image files
symfony/browser-kit                v6.1.0  v6.1.0  Simulates the behavior of a web browser, allowing you to make requests, click on links and submit forms programmatically
symfony/cache                      v5.4.9  v6.1.2  Provides an extended PSR-6, PSR-16 (and tags) implementation
symfony/cache-contracts            v2.5.1  v3.1.1  Generic abstractions related to caching
symfony/config                     v6.1.0  v6.1.0  Helps you find, load, combine, autofill and validate configuration values of any kind
symfony/console                    v5.4.9  v6.1.2  Eases the creation of beautiful and testable command line interfaces
symfony/debug-bundle               v6.1.0  v6.1.0  Provides a tight integration of the Symfony VarDumper component and the ServerLogCommand from MonologBridge into the Symfony full-stack framework
symfony/dependency-injection       v6.1.0  v6.1.2  Allows you to standardize and centralize the way objects are constructed in your application
symfony/deprecation-contracts      v3.1.0  v3.1.1  A generic function and convention to trigger deprecation notices
symfony/dom-crawler                v6.1.0  v6.1.0  Eases DOM navigation for HTML and XML documents
symfony/dotenv                     v6.1.0  v6.1.0  Registers environment variables from a .env file
symfony/error-handler              v6.1.0  v6.1.0  Provides tools to manage errors and ease debugging PHP code
symfony/event-dispatcher           v6.1.0  v6.1.0  Provides tools that allow your application components to communicate with each other by dispatching events and listening to them
symfony/event-dispatcher-contracts v3.1.0  v3.1.1  Generic abstractions related to dispatching event
symfony/expression-language        v5.4.8  v6.1.2  Provides an engine that can compile and evaluate expressions
symfony/filesystem                 v6.1.0  v6.1.0  Provides basic utilities for the filesystem
symfony/finder                     v6.1.0  v6.1.0  Finds files and directories via an intuitive fluent interface
symfony/flex                       v1.19.1 v2.2.2  Composer plugin for Symfony
symfony/form                       v6.1.1  v6.1.2  Allows to easily create, process and reuse HTML forms
[...]

PR #279, opened 2022-06-30
Fix handling non-array return from route provider

andrew-demb:
| Q             | A
| ------------- | ---
| Branch?       | "master" (it is not a new feature, but there is no release branch for 3.0)
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #278
| License       | MIT
| Doc PR        | n/a


dbu (approved):
good catch. we should start using phpstan on this repository ;-)

(if you want to introduce that to the build, i would appreciate it)

